PR #97, opened 2025-08-07
[apps] Add streaming functionality

MehakBindra:
Added streaming functionality in `app/http_stream.py`
Unit Tests: `app/tests/test_streaming.py`
E2E Test : `tests/stream`

Added asynchronous retry utility `utils\retry.py` based on TS and C# for use in streaming.

Made changes to `http_plugin.py` and `app.py` to `create_stream`
Made changes to `activity_context.py` to expose streamer

UPDATES: Mostly made minor changes related to comments
Major Changes:  
- Jitter in `retry.py`
- Using `Timeout` in `http_stream.py`  so as to only cancel a task if it has started.

MORE UPDATES: 
- In `clients/conversation/activity.py`, instead of checking for "typing" activity specifically, check if ID exists, else set to placeholder value.
- In `apps/app_process.py`, register"chunk" and "close" event handlers and close stream after middleware chain.
- In `apps/auth/service_token_validator.py`, add leeway of 30s to validate token with bot framework.
- In `common/events/event_emitter.py`, make once_wrapper async if handler is async.

https://github.com/user-attachments/assets/7753bd2b-449d-4b64-9f08-e73766e10fe0


packages/api/src/microsoft/teams/api/activities/typing.py
```diff
@@ -44,7 +44,14 @@ def add_stream_update(self, sequence: int = 1) -> Self:
             self.channel_data = ChannelData()
 
         self.channel_data.stream_id = self.id
-        self.channel_data.stream_type = "streaming"
         self.channel_data.stream_sequence = sequence
-
-        return self.add_entity(StreamInfoEntity(stream_id=self.id, stream_type="streaming", stream_sequence=sequence))
+        if self.channel_data.stream_type is None:
```
lilyydu, line 48: I believe `stream_sequence` can also be null so we should add a check here as well!

packages/api/src/microsoft/teams/api/clients/conversation/activity.py
```diff
@@ -39,16 +39,13 @@ async def create(self, conversation_id: str, activity: ActivityParams) -> SentAc
...
             f"{self.service_url}/v3/conversations/{conversation_id}/activities",
             json=activity.model_dump(by_alias=True, exclude_none=True),
         )
 
-        if activity.type == "typing":
-            # NOTE: Service returns empty response
-            return SentActivity(id="DO_NOT_USE_PLACEHOLDER_TYPING_ID", activity_params=activity)
-
-        id = response.json()["id"]
+        id = response.json().get("id", "DO_NOT_USE_PLACEHOLDER_TYPING_ID")
```
lilyydu, line 48: could you keep a note that the service returns an empty response for a non-streaming typing activity?

packages/apps/src/microsoft/teams/apps/app_process.py
```diff
@@ -121,15 +122,32 @@ async def updated_send(
 
             self.logger.debug("Calling on_activity_sent for plugins")
 
-            ref = conversation_ref or activityCtx.conversation_ref
```
lilyydu, line 124: this line needs to stay inside the inner function

packages/apps/src/microsoft/teams/apps/http_stream.py
```diff
@@ -0,0 +1,259 @@
...
+        Send status updates before emitting (ex. "Thinking...").
+
+        Args:
+            text: The status text to send.
+        """
+        self.emit(TypingActivityInput().with_text(text).with_channel_data(ChannelData(stream_type="informative")))
+
+    async def close(self) -> Optional[SentActivity]:
+        # wait for lock to be free
+        if self._result is not None:
```
lilyydu, line 132: nit: can do `if self._result`

packages/apps/src/microsoft/teams/apps/http_stream.py
```diff
@@ -0,0 +1,259 @@
...
+            self._logger.debug("stream already closed with result")
+            return self._result
+
+        if self._index == 1 and not self._queue and not self._lock.locked():
+            self._logger.debug("stream has no content to send, returning None")
```
lilyydu, line 137: should our index start at 0 instead?

packages/apps/src/microsoft/teams/apps/http_stream.py
```diff
@@ -0,0 +1,259 @@
...
+            return None
+
+        # Wait until _id is set and queue is empty
+        if not self._id:
```
lilyydu, line 141: should we do a `while` here? (comparing to TS)

packages/apps/src/microsoft/teams/apps/http_stream.py
```diff
@@ -0,0 +1,259 @@
...
+            self._logger.debug("waiting for ID to be set")
+            await self._id_set_event.wait()
+
+        while self._queue:
+            self._logger.debug("waiting for queue to be empty...")
+            await self._queue_empty_event.wait()
+
+        if self._text == "" and self._attachments == []:
+            self._text = "Stream completed without content"
```
lilyydu, line 150: we don't want to return this as the text - should log this instead + return

packages/apps/src/microsoft/teams/apps/http_stream.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        Send or update an activity to the Teams conversation.
+
+        Args:
+            activity: The activity to send.
+        """
+        to_send.from_ = self._ref.bot
+        to_send.conversation = self._ref.conversation
+
+        if to_send.id and not any(e.type == "streaminfo" for e in (to_send.entities or [])):
```
lilyydu, line 256: nit: this line reads a little cryptic, are you able to break it down more

packages/apps/src/microsoft/teams/apps/utils/retry.py
```diff
@@ -0,0 +1,101 @@
...
+        self.delay = delay
+        self.max_delay = max_delay
+        self.jitter_type: JitterType = jitter_type
+        self.logger = logger.getChild("@teams/retry") if logger else ConsoleLogger().create_logger("@teams/retry")
+        self.previous_delay = previous_delay
+
+
+def _apply_jitter(delay: float, jitter_type: JitterType, previous_delay: Optional[float] = None) -> float:
+    """Apply jitter to the delay to prevent thundering herd problems."""
+    if jitter_type == "none":
```
lilyydu, line 39: nit: can use the literal, `JitterType.None `

packages/apps/tests/test_streaming.py
```diff
@@ -0,0 +1,161 @@
...
+    ApiClient,
+    ConversationAccount,
+    ConversationReference,
+    SentActivity,
+    TypingActivityInput,
+)
+from microsoft.teams.apps import HttpStream
+
+
+class TestHttpStream:
```
lilyydu, line 23: nit: renaming file name / test class name to match

packages/apps/tests/test_streaming.py
```diff
@@ -0,0 +1,161 @@
...
+            activity_id="test-activity",
+            channel_id="msteams",
+        )
+
+    @pytest.fixture
+    def http_stream(self, mock_api_client, conversation_reference, mock_logger):
+        return HttpStream(mock_api_client, conversation_reference, mock_logger)
+
+    @pytest.mark.asyncio
+    async def test_stream_emit_message_flushes_after_500ms(self, mock_api_client, conversation_reference, mock_logger):
```
lilyydu, line 66: can we add into one of the test cases to check a sequence of `.update` and `.emit`

lilyydu (commented):
Nice job incorporating!

packages/api/src/microsoft/teams/api/activities/typing.py
MehakBindra: regardless of whether or not self.channel_data.stream_sequence is null or not, it should be set to what's incoming in the function

packages/apps/src/microsoft/teams/apps/http_stream.py
MehakBindra: https://github.com/microsoft/teams.net/blob/32093fb5f2bb3f60ef49678704e42742df35846f/Libraries/Microsoft.Teams.Plugins/Microsoft.Teams.Plugins.AspNetCore/AspNetCorePlugin.Stream.cs#L25 Keeping consistent with C#, First stream also has to be index=1

packages/apps/src/microsoft/teams/apps/utils/retry.py
MehakBindra: I think I can only do that for Enum, not Literal

packages/api/src/microsoft/teams/api/clients/conversation/activity.py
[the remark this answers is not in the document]
MehakBindra: Actually, even for message activity sent on close_stream , we get empty response back . I'll update the placeholder and add a note

packages/api/src/microsoft/teams/api/activities/typing.py
lilyydu: ah it looks like it defaults on L41, all g!

packages/apps/src/microsoft/teams/apps/app_process.py
MehakBindra: Got it !

packages/apps/src/microsoft/teams/apps/http_stream.py
MehakBindra: would prefer to keep it all inline

packages/api/src/microsoft/teams/api/clients/conversation/activity.py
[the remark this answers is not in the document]
heyitsaamir: @lilyydu - if this is true, should we keep id as nullable? 
@MehakBindra you're saying a "create" call leads to empty responses in the final message too? And that "create" call doesn't have an id in its inputparam, correct?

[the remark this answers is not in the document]
MehakBindra: Thanks @lilyydu for the link, yeah basically a create call with no id but  
`"entities":[
    {
      "type": "streaminfo",
      "streamId": "a-0000l", // // (required) must be present for any subsequent request after the first chunk.
      "streamType": "final", // (required) final is only allowed for the last message of the streaming.
    }
  ],`  will have a 202 0K{ } response

packages/apps/src/microsoft/teams/apps/app_process.py
```diff
@@ -121,15 +122,32 @@ async def updated_send(
...
+
+        async def handle_close(close_activity: SentActivity):
+            if self.event_manager:
+                await self.event_manager.on_activity_sent(
+                    sender,
+                    ActivitySentEvent(sender=sender, activity=close_activity, conversation_ref=ref),
+                    plugins=plugins,
+                )
+
+        activityCtx.stream.events.on("chunk", handle_chunk)
```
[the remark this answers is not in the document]
MehakBindra: thanks ! resolving

PR #98, opened 2025-08-08
Add augmentable activity context

heyitsaamir:
This is a similar draft for being able to create augmentable activity context (See https://github.com/microsoft/teams.ts/pull/288). Here is how it actually looks like in practice:

```py


class Foo:
    bar: str


class FooPlugin(Plugin[Foo]):
    def on_activity(self):
        return Foo()


app = App(AppOptions(plugins=[FooPlugin()]))

@app.on_message
async def handle_message(ctx: ActivityContext[MessageActivity]):
   foo_ctx = ctx.get_plugin_context(FooPlugin)
   foo_ctx.bar # this is correctly str
```

What this means is that we can build plugins like a GraphPlugin which will then give us access to graph_ctx.user_graph, for example. And it gives plugins additonal flexibility to add custom things to the activity context as desired.

In typescript, the nice thing is that we can just do activityContext.bar, and it'll be correctly inferred. But python's type system is not nearly as powerful as typescript, so we have to explicitly ask ActivityContext the type for the plugin (hence `foo_ctx = ctx.get_plugin_context(FooPlugin)`) so we get the right types.

### DIFF
--- a/packages/app/src/microsoft/teams/app/app.py
+++ b/packages/app/src/microsoft/teams/app/app.py
@@ -8,10 +8,11 @@ import importlib.metadata
 import os
 from dataclasses import dataclass
 from logging import Logger
-from typing import Any, Callable, Dict, List, Optional, TypeVar, Union, cast, overload
+from typing import Any, Callable, Dict, List, Optional, Type, TypeVar, Union, cast, overload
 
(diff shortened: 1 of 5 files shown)


heyitsaamir:
Removing from draft. But keeping the branch.

PR #114, opened 2025-08-22
[apps & tests/echo] fix json parsing, id access, await, static access

lilyydu:
- using `ActivityTypeAdapter` to parse the JSON body into the correct `Activity` type
- used asyncio.gather to start all the plugins
- removed unnecessary conditional callables
- fix attribute error on `activity.id`
- use placeholder loud value for id on typing activities
- fix await on the `on_activity_event `


packages/apps/src/microsoft/teams/apps/http_plugin.py
```diff
@@ -249,12 +252,15 @@ async def _handle_activity_request(self, request: Request) -> Any:
         response_future = asyncio.get_event_loop().create_future()
         self.pending[activity_id] = response_future
 
+        type_adapter: TypeAdapter[Activity] = TypeAdapter(Activity)
```
heyitsaamir, line 255: Use `ActivityTypeAdapter`. That's what we were using before

packages/apps/src/microsoft/teams/apps/http_plugin.py
```diff
@@ -249,12 +252,15 @@ async def _handle_activity_request(self, request: Request) -> Any:
...
+        activity = type_adapter.validate_python(body)
+
         # Fire activity processing via callback
         if callable(self.on_activity_event):
```
heyitsaamir, line 259: why are we checking for on_activity_event being a callable? we don't seem to use it here

packages/apps/src/microsoft/teams/apps/http_plugin.py
```diff
@@ -213,7 +213,10 @@ async def _process_activity(self, activity: Activity, activity_id: str, token: T
         try:
             if callable(self.on_activity):
                 event = ActivityEvent(activity=activity, sender=self, token=token)
-                self.on_activity_event(event)
+                if asyncio.iscoroutinefunction(self.on_activity_event):
+                    await self.on_activity_event(event)
+                else:
+                    self.on_activity_event(event)
```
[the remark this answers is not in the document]
heyitsaamir: Agree with copilot...

tests/echo/src/main.py
```diff
@@ -25,7 +24,6 @@ async def handle_message(ctx: ActivityContext[MessageActivity]):
     """Handle message activities using the new generated handler system."""
     print(f"[GENERATED onMessage] Message received: {ctx.activity.text}")
     print(f"[GENERATED onMessage] From: {ctx.activity.from_}")
-    await ctx.reply(TypingActivityInput())
```
heyitsaamir, line 28: Let's put this back once you're satisfied with this working

packages/apps/src/microsoft/teams/apps/app.py
```diff
@@ -206,22 +206,21 @@ async def start(self, port: Optional[int] = None) -> None:
...
             async def on_http_ready() -> None:
                 self.log.info("Teams app started successfully")
                 assert self._port is not None, "Port must be set before emitting start event"
                 self._events.emit("start", StartEvent(port=self._port))
 
             self.http.on_ready_callback = on_http_ready
-            event = PluginStartEvent(port=self._port)
-            await self.http.on_start(event)
+
+            tasks: List[Awaitable[Any]] = []
```
heyitsaamir, line 217: should this be in app.plguins?

packages/apps/src/microsoft/teams/apps/app.py
```diff
@@ -206,22 +206,21 @@ async def start(self, port: Optional[int] = None) -> None:
...
+            for plugin in self.plugins:
+                is_callable = hasattr(plugin, "on_start") and callable(plugin.on_start)
+                if is_callable:
+                    event = PluginStartEvent(port=self._port)
```
heyitsaamir, line 221: NIT: can create this outside the loop?

packages/devtools/src/microsoft/teams/devtools/devtools_plugin.py
```diff
@@ -49,7 +49,7 @@ class DevToolsPlugin(Sender):
     logger: Annotated[Logger, LoggerDependencyOptions()]
     id: Annotated[Optional[TokenProtocol], DependencyMetadata(optional=True)]
     name: Annotated[Optional[TokenProtocol], DependencyMetadata(optional=True)]
```
heyitsaamir, line 51: One potential nicer thing could be this:
1. When registering a particular dep, we can register with "is_optional".
2. When we're parsing this, we can check if it's Annotated[Optional[...]. If it's not, we can warn that the plugin is treating an optional plugin as non-optional.

packages/devtools/src/microsoft/teams/devtools/devtools_plugin.py
```diff
@@ -87,7 +87,18 @@ async def websocket_endpoint(websocket: WebSocket):  # type: ignore
...
+        @self.app.get("/devtools/{path:path}")
+        async def custom_path(request: Request, path: str):  # type: ignore
+            file_path = os.path.join(dist, path)
+            if os.path.isfile(file_path):
+                return FileResponse(file_path)
+            return FileResponse(os.path.join(dist, "index.html"))
+
+        @self.app.get("/devtools")
+        async def root():  # type: ignore
+            return FileResponse(os.path.join(dist, "index.html"))
```
heyitsaamir, line 101: NIT: why ignore types?

packages/apps/src/microsoft/teams/apps/app.py
lilyydu: i think it makes more sense in app.py cause only the app class should have the power to start the plugins

packages/devtools/src/microsoft/teams/devtools/devtools_plugin.py
lilyydu: pushing ideation/discussion to next PR to unblock Mehak!

PR #119, opened 2025-08-29
Initialize AI package

heyitsaamir:
1. Initializes an AI package (should we call it ai-core?)
2. Initializes some protocols for AI Models
3. Created ChatPrompt class
4. Built the models for Openai/AzureOpenai chat-completions
5. Add ai-test to test functionality
6. (Also added an "Agent" class that's basically ChatPrompt, but more "stateful". ChatPrompt is stateless, but "Agent" isn't). Thinking through if I like this approach or not. 

#### PR Dependency Tree

* **PR #119** 👈
  * **PR #120**
    * **PR #121**
      * **PR #123**


packages/ai/pyproject.toml
```diff
@@ -0,0 +1,26 @@
+[project]
+name = "microsoft-teams-ai"
+version = "0.0.1a2"
```
lilyydu, line 3: update to a3

packages/ai/src/microsoft/teams/ai/agent.py
```diff
@@ -0,0 +1,27 @@
...
+from typing import Any, Awaitable, Callable
+
+from .ai_model import AIModel
+from .chat_prompt import ChatPrompt, ChatSendResult
+from .function import Function
+from .memory import Memory
+from .message import Message
+
+
+class Agent(ChatPrompt):
```
lilyydu, line 15: could we add a desc here that notes this is a stateful vers of ChatPrompt?

packages/openai_ai_model/pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[project]
+name = "microsoft-teams-openai-ai-model"
+version = "0.0.1-alpha.1"
```
lilyydu, line 3: bump vers here also

packages/openai_ai_model/src/microsoft/teams/openai_ai_model/ai_model.py
```diff
@@ -0,0 +1,298 @@
...
+                content += delta.content
+                if on_chunk:
+                    await on_chunk(delta.content)
+
+            # Handle tool calls streaming
+            if delta.tool_calls:
+                for call_delta in delta.tool_calls:
+                    # Ensure we have enough slots for this index
+                    while len(tool_calls_data) <= call_delta.index:
+                        tool_calls_data.append({"id": "", "name": "", "arguments_str": ""})
```
lilyydu, line 197: nit: can we use `ToolCallData` dict defined above

packages/ai/src/microsoft/teams/ai/agent.py
heyitsaamir: Yes!

packages/openai_ai_model/src/microsoft/teams/openai_ai_model/ai_model.py
heyitsaamir: We are aren't we?

packages/ai/src/microsoft/teams/ai/agent.py
```diff
@@ -0,0 +1,32 @@
...
+        self.memory = memory or ListMemory()
+
+    async def send(
+        self,
+        input: str | Message,
+        *,
+        memory: Memory | None = None,
+        on_chunk: Callable[[str], Awaitable[None]] | Callable[[str], None] | None = None,
+    ) -> ChatSendResult:
+        return await super().send(input, memory=memory or self.memory, on_chunk=on_chunk)
```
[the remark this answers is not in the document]
heyitsaamir: I think i prefer this approach because "send" shouldn't have this side effect of modifying the agent's memory. if the user decides to pass a custom memory, we just honor it for that call, but it's not really expected to be passed in.

PR #123, opened 2025-09-03
Add MCP Client Plugin

heyitsaamir:


packages/mcp/README.md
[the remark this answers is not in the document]
heyitsaamir: hmmm, good point. i think that's a good idea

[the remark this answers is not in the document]
heyitsaamir: Actually, idk, the external thing seems like overkill to me now. I say let's keep it in packages. Simpler to maintain. Thoughts?

packages/openai/src/microsoft/teams/openai/responses_chat_model.py
```diff
@@ -23,6 +23,10 @@
 from pydantic import BaseModel
 
 from openai import NOT_GIVEN
+from openai.lib._pydantic import (
+    _ensure_strict_json_schema,  # pyright: ignore [reportPrivateUsage]
```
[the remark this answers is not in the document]
heyitsaamir: It's looked down upon. Everything's public in python :)

packages/mcp/src/microsoft/teams/mcp/ai_plugin.py
```diff
@@ -0,0 +1,262 @@
...
+import logging
+import time
+from typing import Any, Awaitable, Callable, Dict, List, Optional, Tuple, Union
+
+from microsoft.teams.ai.function import Function
+from microsoft.teams.ai.plugin import BaseAIPlugin
+from microsoft.teams.common.logging import ConsoleLogger
+from pydantic import BaseModel
+
+from mcp import ClientSession
```
lilyydu, line 17: nit: move external imports to top (mb we can add this into ruff?)

packages/mcp/src/microsoft/teams/mcp/ai_plugin.py
```diff
@@ -0,0 +1,262 @@
...
+from mcp.types import TextContent
+
+from .transport import create_transport
+
+REFETCH_TIMEOUT_MS = 24 * 60 * 60 * 1000  # 1 day
+
+
+class McpToolDetails(BaseModel):
```
lilyydu, line 25: nit: could we have this in a separate file - same w cached value? separation of concerns

packages/mcp/src/microsoft/teams/mcp/ai_plugin.py
heyitsaamir: Let me do this in a separate PR. I'll run import reordering across all files

heyitsaamir: https://docs.astral.sh/ruff/settings/#lint_isort_sections going to use this (just so i remember)

packages/mcp/src/microsoft/teams/mcp/ai_plugin.py
heyitsaamir: Moved em to their own model files

packages/mcp/README.md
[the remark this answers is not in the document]
heyitsaamir: Btw if you still feel strongly that we should, let's do it as a separate PR because it'll involve a bunch of changes unrelated to mcp :)

PR #126, opened 2025-09-04
Create MCP Server

heyitsaamir:


packages/apps/src/microsoft/teams/apps/plugins/metadata.py
```diff
@@ -20,10 +20,15 @@ class PluginOptions:
...
+T = TypeVar("T")
+
+
+def Plugin(
+    name: Optional[str] = None, version: Optional[str] = None, description: Optional[str] = None
+) -> Callable[[Type[T]], Type[T]]:
     """Turns any class into a plugin using the decorator pattern."""
 
-    def decorator(cls: Type[Any]) -> Any:
+    def decorator(cls: Type[T]) -> Type[T]:
```
heyitsaamir, line 31: Had to use generics here because without it, this definition was setting the class to be "Any" which was losing all its type checking goodness.

packages/apps/src/microsoft/teams/apps/http_plugin.py
```diff
@@ -89,7 +93,16 @@ async def lifespan(_app: FastAPI) -> AsyncGenerator[None, None]:
             if self._on_stopped_callback:
                 await self._on_stopped_callback()
 
-        self.app = FastAPI(lifespan=lifespan)
+        @asynccontextmanager
+        async def combined_lifespan(app: Any):
```
heyitsaamir, line 97: had to add this because other plugins may also want in on the lifespan in fastapi, and it this exposes it to other plugins and combines it when the app is run

PR #127, opened 2025-09-04
Unpack App Options

heyitsaamir:
Changes AppOptions to a TypedDict, and uses `Unpack` in `App` constructor. This allows us to do a simple App(client_id="Foo", ...) pattern, which is much cleaner. Fans of the old approach may also do App(**AppOptions(...)).

#### PR Dependency Tree

* **PR #127** 👈

### DIFF
--- a/packages/apps/src/microsoft/teams/apps/app.py
+++ b/packages/apps/src/microsoft/teams/apps/app.py
@@ -6,7 +6,6 @@ Licensed under the MIT License.
 import asyncio
 import importlib.metadata
 import os
-from dataclasses import asdict
 from logging import Logger
 from typing import Any, Awaitable, Callable, List, Optional, TypeVar, Union, Unpack, cast, overload
 
@@ -42,7 +41,7 @@ from .events import (
 )
 from .graph_token_manager import GraphTokenManager
 from .http_plugin import HttpPlugin
-from .options import AppOptions, AppOptionsDefaults
+from .options import AppOptions, merge_app_options_with_defaults
 from .plugins import PluginBase, PluginStartEvent, get_metadata
 from .routing import ActivityHandlerMixin, ActivityRouter
 
@@ -62,8 +61,7 @@ class App(ActivityHandlerMixin):
     """
 
     def __init__(self, **options: Unpack[AppOptions]):
-        defaults = asdict(AppOptionsDefaults())
-        self.options = cast(AppOptions, {**defaults, **options})
+        self.options = merge_app_options_with_defaults(**options)
 
         self.log = self.options.get("logger") or ConsoleLogger().create_logger("@teams/app")
         self.storage = self.options.get("storage") or LocalStorage()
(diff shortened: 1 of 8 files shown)


packages/apps/src/microsoft/teams/apps/app.py
```diff
@@ -60,11 +61,12 @@ class App(ActivityHandlerMixin):
...
 
-    def __init__(self, options: Optional[AppOptions] = None):
-        self.options = options or AppOptions()
+    def __init__(self, **options: Unpack[AppOptions]):
+        defaults = asdict(AppOptionsDefaults())
+        self.options = cast(AppOptions, {**defaults, **options})
 
-        self.log = self.options.logger or ConsoleLogger().create_logger("@teams/app")
-        self.storage = self.options.storage or LocalStorage()
+        self.log = self.options.get("logger") or ConsoleLogger().create_logger("@teams/app")
```
lilyydu, line 68: will typed dict throw an error if the key DNE?

packages/apps/src/microsoft/teams/apps/options.py
```diff
@@ -5,27 +5,35 @@
...
+from typing import Any, List, Optional, TypedDict
 
 from microsoft.teams.common.storage import Storage
 
 from .plugins import PluginBase
 
 
-@dataclass
-class AppOptions:
+class AppOptions(TypedDict, total=False):
```
lilyydu, line 15: by setting total=False, does it still check that all req fields are being passed in?

packages/apps/src/microsoft/teams/apps/options.py
```diff
@@ -5,27 +5,35 @@
...
+    storage: Optional[Storage[str, Any]]
+    plugins: List[PluginBase]
+    enable_token_validation: bool
 
     # Oauth
+    default_connection_name: str
+
+
+@dataclass
+class AppOptionsDefaults:
```
lilyydu, line 34: thoughts on a factory method instead for defaults, so we don't need to maintain the dict/class?
ex - https://stackoverflow.com/a/76251992

lilyydu: ^ this might also help with type checking cause a `TypedDict` is not enforced at runtime

heyitsaamir: Like it! Let me update

packages/apps/src/microsoft/teams/apps/options.py
heyitsaamir: No it doesn't (which is what we want, since everything is optional)
> This indicates that all keys defined within the TypedDict are optional by default. Instances of the TypedDict can be created without including all, or even any, of the declared keys, and a type checker will not raise an error for missing keys.

heyitsaamir: But i switched to using NotRequired for now. I think it's cleaner

PR #128, opened 2025-09-04
Merge the activity sent with response in http stream before passing it back to event manager in app 

MehakBindra:
Screenshot 2025-09-04 162334

Change: `return res`  ->  `return SentActivity.merge(to_send, res)`  in `_send` method of `http_stream.py`.

This gets sent to `app.event_manager.on_activity_sent(....)` which routes it to `devtools_plugin.on_activity_sent(...)` where we emit activity to web socket `await self.emit_activity_to_sockets(activity)`.

Without this change, only the content inside activity_params is present 
image


heyitsaamir (approved):
Can you update the PR description? Those become documentation :)

PR #139, opened 2025-09-09
Add app.use middleware

### LINKED ISSUE #135: [py] add in missing app.use middleware
add to the router, with lambda _ctx: true selector
TS equivalent - https://microsoft.github.io/teams-ai/typescript/in-depth-guides/observability/middleware

zion-off:
Closes #135 and follows the same pattern as the [TypeScript SDK](https://microsoft.github.io/teams-ai/typescript/in-depth-guides/observability/middleware/)

### Usage:

Now you can do something like

```python
async def your_middleware(ctx):
    # your logic
    await ctx.next()
    # your logic

app.use(your_middleware)
```

and it will run every middleware in sequence before passing it on to the handlers

### DIFF
--- a/packages/apps/src/microsoft/teams/apps/app.py
+++ b/packages/apps/src/microsoft/teams/apps/app.py
@@ -13,6 +13,7 @@ from dependency_injector import providers
 from dotenv import find_dotenv, load_dotenv
 from microsoft.teams.api import (
     Account,
+    ActivityBase,
     ActivityParams,
     ApiClient,
     ClientCredentials,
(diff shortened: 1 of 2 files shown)


zion-off:
@heyitsaamir i've overwritten my commit with the changes to make it more aligned with the ts sdk; looks much simpler now. does this look good?

thanks so much for reviewing!

packages/apps/src/microsoft/teams/apps/app.py
```diff
@@ -284,6 +286,10 @@ async def send(self, conversation_id: str, activity: str | ActivityParams | Adap
 
         return await self.http.send(activity, conversation_ref)
 
+    def use(self, middleware: Callable[[ActivityContext[ActivityBase]], Awaitable[None]]) -> None:
```
heyitsaamir, line 289: Great! Could you add a test?

heyitsaamir (approved):

PR #172, opened 2025-10-14
[Bug] conversationId in tab sample should be camelCase as it is consumed by React App

MehakBindra:
1. Changed conversationId in tab sample function response back to camelCase.
2. `self._conversation_id` - > `self._resolved_conversation_id` in `function_context.py` ( consistent with the field in ClientContext).
3. Default the `entra_token_validator` to None in app class and check in jwt validation if it exists. If not, throw error.

### DIFF
--- a/packages/apps/src/microsoft/teams/apps/auth/remote_function_jwt_middleware.py
+++ b/packages/apps/src/microsoft/teams/apps/auth/remote_function_jwt_middleware.py
@@ -20,7 +20,7 @@ def require_fields(fields: Dict[str, Optional[Any]], context: str, logger: Logge
         raise HTTPException(status_code=401, detail=message)
 
 
-def remote_function_jwt_validation(logger: Logger, entra_token_validator: Optional[TokenValidator] = None):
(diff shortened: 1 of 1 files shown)


packages/apps/src/microsoft/teams/apps/auth/remote_function_jwt_middleware.py
```diff
@@ -20,7 +20,7 @@ def require_fields(fields: Dict[str, Optional[Any]], context: str, logger: Logge
         raise HTTPException(status_code=401, detail=message)
 
 
-def remote_function_jwt_validation(entra_token_validator: TokenValidator, logger: Logger):
+def remote_function_jwt_validation(logger: Logger, entra_token_validator: Optional[TokenValidator] = None):
```
heyitsaamir: if we're just gonna throw below, why make it optional?

heyitsaamir: So let's not default to None?

PR #190, opened 2025-10-29
[feat] Add BotBuilder Plugin package, sample and tests

MehakBindra:
- Added `BotBuilderPlugin` (overrides http plugin's `on_activity_request`)
- Refactored `HttpPlugin` to easily extend to botbuilder without code duplication
- Added a test sample
- Added unit tests
- Added `HttpPluginOptions` to be able to extend easily
- Moved jwt validation initialization to on init in http plugin to leverage DI for credentials 

Screenshot 2025-10-28 171527

### DIFF
--- a/packages/apps/src/microsoft/teams/apps/http_plugin.py
+++ b/packages/apps/src/microsoft/teams/apps/http_plugin.py
@@ -17,7 +17,6 @@ from typing import (
     Callable,
     Dict,
     Optional,
-    Required,
     TypedDict,
     Union,
     Unpack,
@@ -36,6 +35,7 @@ from microsoft.teams.api import (
     SentActivity,
     TokenProtocol,
 )
+from microsoft.teams.api.auth.credentials import Credentials
 from microsoft.teams.apps.http_stream import HttpStream
 from microsoft.teams.common.http import Client, ClientOptions, Token
 from microsoft.teams.common.logging import ConsoleLogger
@@ -63,7 +63,6 @@ version = importlib.metadata.version("microsoft-teams-apps")
 class HttpPluginOptions(TypedDict, total=False):
     """Options for configuring the HTTP plugin."""
 
-    app_id: Required[Optional[str]]  # must be present, but can be None
     logger: Logger
     skip_auth: bool
     server_factory: Callable[[FastAPI], uvicorn.Server]
@@ -76,6 +75,7 @@ class HttpPlugin(Sender):
     """
 
     logger: Annotated[Logger, LoggerDependencyOptions()]
+    credentials: Annotated[Optional[Credentials], DependencyMetadata(optional=True)]
 
     on_error_event: Annotated[Callable[[ErrorEvent], None], EventMetadata(name="error")]
     on_activity_event: Annotated[Callable[[ActivityEvent], None], EventMetadata(name="activity")]
@@ -89,7 +89,6 @@ class HttpPlugin(Sender):
     def __init__(self, **options: Unpack[HttpPluginOptions]):
         """
         Args:
-            app_id: Optional Microsoft App ID.
             logger: Optional logger.
             skip_auth: Whether to skip JWT validation.
             server_factory: Optional function that takes an ASGI app
@@ -106,6 +105,7 @@ class HttpPlugin(Sender):
         super().__init__()
         self.logger = options.get("logger") or ConsoleLogger().create_logger("@teams/http-plugin")
         self._port: Optional[int] = None
+        self._skip_auth: bool = options.get("skip_auth", False)
         self._server: Optional[uvicorn.Server] = None
         self._on_ready_callback: Optional[Callable[[], Awaitable[None]]] = None
         self._on_stopped_callback: Optional[Callable[[], Awaitable[None]]] = None
@@ -146,15 +146,6 @@ class HttpPlugin(Sender):
                     "server_factory must return a uvicorn.Server configured with the provided FastAPI app instance."
                 )
 
-        # Add JWT validation middleware
-        app_id = options.get("app_id")
-        skip_auth = options.get("skip_auth", False)
-        if app_id and not skip_auth:
-            jwt_middleware = create_jwt_validation_middleware(
-                app_id=app_id, logger=self.logger, paths=["/api/messages"]
-            )
-            self.app.middleware("http")(jwt_middleware)
-
         # Expose FastAPI routing methods (like TypeScript exposes Express methods)
         self.get = self.app.get
         self.post = self.app.post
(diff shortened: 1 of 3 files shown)


packages/apps/src/microsoft/teams/apps/http_plugin.py
```diff
@@ -306,38 +306,51 @@ async def _handle_activity_request(self, request: Request) -> Any:
 
         return result
 
+    def _handle_activity_response(self, response: Response, result: Any) -> Any:
```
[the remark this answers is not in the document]
heyitsaamir: This will need be async

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+
+class BotBuilderPluginOptions(TypedDict, total=False):
+    """Options for configuring the BotBuilder plugin."""
+
+    skip_auth: bool
+    handler: ActivityHandler
+    adapter: CloudAdapter
+
+
+@Plugin(name="http", version=version, description="BotBuilder plugin for Microsoft Bot Framework integration")
```
lilyydu: plugin name should be botbuilder/what we have in TS/.NET

MehakBindra: plugin name is http in TS as well, it replaces the existing HTTP plugin and overrides it's behavior for handling activities

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    BotBuilder plugin that provides Microsoft Bot Framework integration.
+    """
+
+    # Dependency injections
+    logger: Annotated[Logger, LoggerDependencyOptions()]
+    credentials: Annotated[Optional[Credentials], DependencyMetadata(optional=True)]
+    client: Annotated[Client, DependencyMetadata()]
+
+    bot_token: Annotated[Optional[Callable[[], TokenProtocol]], DependencyMetadata(optional=True)]
+    graph_token: Annotated[Optional[Callable[[], TokenProtocol]], DependencyMetadata(optional=True)]
```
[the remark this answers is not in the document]
MehakBindra: You're right, not needed !

tests/botbuilder/src/main.py
```diff
@@ -0,0 +1,39 @@
...
+from microsoft.teams.devtools import DevToolsPlugin
+
+
+class MyActivityHandler(ActivityHandler):
+    async def on_message_activity(self, turn_context: TurnContext):
+        print("Message activity received.")
+        await turn_context.send_activity("hi from botbuilder...")
+
+
+handler = MyActivityHandler()
```
heyitsaamir: Hm, this isn't the best example of how one could migrate. Take a look at a typical bot builder sample (like an echo bot). Use that as an example that can be passed into BotBuildPlugin

tests/botbuilder/src/main.py
```diff
@@ -0,0 +1,39 @@
...
+
+app = App(
+    plugins=[
+        BotBuilderPlugin(app_id=os.getenv("CLIENT_ID"), handler=handler),
```
heyitsaamir: why does app id need to be passed in?

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
lilyydu: ooo good to know, maybe we should call it botbuilder-http cause its not very explicit?

packages/apps/src/microsoft/teams/apps/http_plugin.py
```diff
@@ -337,38 +356,51 @@ async def _handle_activity_request(self, request: Request) -> Any:
...
+        if body is not None:
+            self.logger.debug(f"Returning body {body}")
+            return body
+        self.logger.debug("Returning empty body")
+        return response
+
+    async def on_activity_request(self, request: Request, response: Response) -> Any:
+        """Handle incoming Teams activity."""
+        # Process the activity (token validation handled by middleware)
+        result = await self._handle_activity_request(request)
```
lilyydu: I think this is the point that Aamir was saying about async - we're awaiting here

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
```diff
@@ -0,0 +1,140 @@
+# pyright: reportMissingTypeStubs=false, reportUnknownMemberType=false
```
lilyydu: are we able to work around this disablement?

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
```diff
@@ -0,0 +1,148 @@
...
+    def __init__(self, **options: Unpack[BotBuilderPluginOptions]):
+        """
+        Initialize the BotBuilder plugin.
+
+        Args:
+            options: Configuration options for the plugin
+        """
+        self.options = options
+        super().__init__(
+            app_id=None,
```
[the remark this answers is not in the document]
MehakBindra: This is outdated Aamir, Now BotBuilder and Http have same behavior. Explicitly pass in app id during init , even if none.

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
```diff
@@ -0,0 +1,140 @@
...
+)
+from botbuilder.schema import Activity
+
+version = importlib.metadata.version("microsoft-teams-botbuilder")
+
+# Constants for app types
+SINGLE_TENANT = "singletenant"
+MULTI_TENANT = "multitenant"
+
+
```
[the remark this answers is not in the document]
MehakBindra: will do

packages/botbuilder/src/microsoft/teams/botbuilder/botbuilder_plugin.py
MehakBindra: I don't think we can do that, Lily. The application explicitly checks for the "http" plugin name, and if it doesn't find one, it creates a default HTTP plugin instance. In our case, the BotBuilder plugin is already acting as the HTTP layer

if meta.name == "http":
    http_plugin = plugin

lilyydu: ahh.. thats true, good catch!

tests/botbuilder/src/main.py
heyitsaamir: https://github.com/microsoft/BotBuilder-Samples/blob/main/samples/python/02.echo-bot. Try mimicking the pattern here. So if someone looks at this code, they can find similar patterns in their code to be able to migrate it successfully

MehakBindra: Addressed by adding adapter as well

heyitsaamir (commented):
Looks good! 
Asking for a comment, and I still don't fully understand the app_id thing. Why do we need it?

tests/botbuilder/src/main.py
```diff
@@ -0,0 +1,63 @@
...
+            value=f"{error}",
+            value_type="https://www.botframework.com/schemas/error",
+        )
+        await context.send_activity(trace_activity)
+
+
+adapter.on_turn_error = on_error
+
+app = App(
+    plugins=[BotBuilderPlugin(app_id=os.getenv("CLIENT_ID"), adapter=adapter, handler=EchoBot()), DevToolsPlugin()]
```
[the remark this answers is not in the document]
MehakBindra: done, moved app id stuff to on init (resolved via DI)

PR #191, opened 2025-10-29
Introduce msal

heyitsaamir:
In this PR, we introduce [MSAL](https://learn.microsoft.com/en-us/entra/msal/python/) as a dependency to take care of authentication for us. We replace the use of bot token client with this library. It still uses secrets by default, but uses MSAL to do it.

Tested the general flows (regular + graph)

#### PR Dependency Tree

* **PR #191** 👈
  * **PR #192**
    * **PR #203**

### DIFF
--- a/packages/apps/src/microsoft/teams/apps/token_manager.py
+++ b/packages/apps/src/microsoft/teams/apps/token_manager.py
@@ -18,6 +18,12 @@ from microsoft.teams.api.auth.credentials import TokenCredentials
 from microsoft.teams.common import ConsoleLogger
 from msal import ConfidentialClientApplication  # pyright: ignore[reportMissingTypeStubs]
 
+BOT_TOKEN_SCOPE = "https://api.botframework.com/.default"
+GRAPH_TOKEN_SCOPE = "https://graph.microsoft.com/.default"
+DEFAULT_TENANT_FOR_BOT_TOKEN = "botframework.com"
+DEFAULT_TENANT_FOR_GRAPH_TOKEN = "common"
+DEFAULT_TOKEN_AUTHORITY = "https://login.microsoftonline.com/{tenant_id}"
+
 
 class TokenManager:
     """Manages authentication tokens for the Teams application."""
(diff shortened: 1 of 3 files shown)


packages/apps/src/microsoft/teams/apps/token_manager.py
```diff
@@ -70,29 +51,53 @@ async def get_graph_token(self, tenant_id: Optional[str] = None, force: bool = F
         Returns:
             The graph token or None if not available
         """
-        if not self._credentials:
-            self._logger.debug("No credentials provided for graph token refresh")
+        return await self._get_token("https://graph.microsoft.com/.default", tenant_id)
+
+    async def _get_token(
+        self, scope: str | list[str], tenant_id: str | None = None, *, caller_name: str | None = None
```
lilyydu: always wondered about this, but is it better to sandwich the * ?

stubs/msal/__init__.pyi
```diff
@@ -0,0 +1,13 @@
+"""Type stubs for msal"""
+
+from typing import Any, Optional
+
+class ConfidentialClientApplication:
+    """MSAL Confidential Client Application"""
+
+    def __init__(
+        self, client_id: str, *, client_credential: Optional[str] = None, authority: Optional[str] = None, **kwargs: Any
```
lilyydu: nit: using Optional here but | in other files, we haven't adopted an official standard but mb we should try going forward

heyitsaamir: I like |

packages/apps/src/microsoft/teams/apps/token_manager.py
heyitsaamir: Hm, not sure tbh. I just think of it as stuff that's essential for the operation of a function vs not. I'll actually remove the default param from tenant_id

packages/apps/src/microsoft/teams/apps/token_manager.py
```diff
@@ -70,29 +51,53 @@ async def get_graph_token(self, tenant_id: Optional[str] = None, force: bool = F
...
+            tenant_id = tenant_id or "botframework.com"
+            token = credentials.token(scope, tenant_id)
+            if isawaitable(token):
+                access_token = await token
+            else:
+                access_token = token
+
+            return JsonWebToken(access_token)
+
+    def _get_msal_client_for_tenant(self, tenant_id: str) -> ConfidentialClientApplication:
```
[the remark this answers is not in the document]
heyitsaamir: Put all the stuff in constants for you. Sounds like we'll need to update the authority to point to a regional endpoint for that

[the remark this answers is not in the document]
heyitsaamir: `DEFAULT_TOKEN_AUTHORITY = "https://login.microsoftonline.com/{tenant_id}"`

PR #192, opened 2025-10-29
Add User Managed Identity support via MSAL

heyitsaamir:
Adds native support for User Managed Identity via MSAL. 

Tested with UMI. 

It's enabled when CLIENT_ID is set and CLIENT_SECRET is not.

| CLIENT_ID | CLIENT_SECRET | MANAGED_IDENTITY_CLIENT_ID | Output |
|-|-|-|-|
| not_set | | | No-Auth |
| set | set | | SecretsAuth |
| set | not_set | | User Managed Identity Auth |
| set | not_set | set (same as CLIENT_ID) | User Managed Identity Auth |
| set | not_set | set (diff from CLIENT_ID) | FIC (user managed identity) |
| set | not_set | "system" | FIC (system identity) |

#### PR Dependency Tree

* **PR #192** 👈
  * **PR #203**

### DIFF
--- a/stubs/msal/__init__.pyi
+++ b/stubs/msal/__init__.pyi
@@ -27,7 +27,7 @@ class ManagedIdentityClient:
 
     def __init__(
         self,
-        managed_identity: SystemAssignedManagedIdentity | UserAssignedManagedIdentity | dict[str, Any],
+        managed_identity: SystemAssignedManagedIdentity | UserAssignedManagedIdentity,
         *,
         http_client: Any,
         token_cache: Any | None = None,


stubs/msal/__init__.pyi
```diff
@@ -9,5 +9,29 @@ class ConfidentialClientApplication:
...
+    """MSAL User Assigned Managed Identity"""
+
+    def __init__(self, *, client_id: str) -> None: ...
+
+class ManagedIdentityClient:
+    """MSAL Managed Identity Client"""
+
+    def __init__(
+        self,
+        managed_identity: SystemAssignedManagedIdentity | UserAssignedManagedIdentity | dict[str, Any],
```
lilyydu: what case would dict[str, Any] handle?

heyitsaamir: Removed it :).

corinagum (approved):
Thanks for the chart. It was helpful.

PR #197, opened 2025-10-31
[Fix] streaming: do not reset timeout for each emit, do not wait forever on close stream

MehakBindra:
Issue : https://github.com/microsoft/teams.ts/issues/374

Main change::
- Emit behavior updated
  - If no flush is scheduled and no timeout is pending, emit flushes immediately.
  - Otherwise, messages are queued and flushed later.

- Flush improvements
  - Explicitly acquire and release the lock to prevent concurrent flushes.
  - If the lock is unavailable, flush returns immediately without blocking.

- Test fixes
  - Added _wait_for_id_and_queue to prevent hanging tests: ensures stream.close() does not wait forever if flush fails.
  - Removed asyncio.sleep delays in tests; replaced with mocked call_later for deterministic scheduling.
  - Added a test to verify concurrent emits are serialized and do not flush at the same time.

- remove `timer.py` (Previously used to differentiate between ongoing tasks and tasks not yet started, which is no longer needed as we now only cancel timeout from flush, not emit)

https://github.com/user-attachments/assets/bee7b959-b2fb-42a2-92a1-83082b2ef97e

### DIFF
--- a/packages/ai/src/microsoft_teams/ai/chat_prompt.py
+++ b/packages/ai/src/microsoft_teams/ai/chat_prompt.py
@@ -185,8 +185,6 @@ class ChatPrompt:
             on_chunk=on_chunk_fn if on_chunk else None,
         )
 
-        print("Response received from model:", response)
-
         current_response = await self._run_after_send_hooks(response)
 
         return ChatSendResult(response=current_response)


packages/apps/tests/test_http_stream.py
```diff
@@ -35,13 +34,12 @@ def mock_api_client(self):
...
 
         client.send_call_count = 0
-        client.send_times = []
         client.sent_activities = []
 
         async def mock_send(activity):
             client.send_call_count += 1
-            client.send_times.append(datetime.now())
             client.sent_activities.append(activity)
+            await asyncio.sleep(0.05)  # Simulate network delay
```
heyitsaamir: why is this necessary? (it'll slow our tests down that's why i'm asking)

packages/apps/tests/test_http_stream.py
```diff
@@ -63,37 +61,45 @@ def http_stream(self, mock_api_client, conversation_reference, mock_logger):
...
+        stream.emit("Fifth message")
+        stream.emit("Sixth message")
+        stream.emit("Seventh message")
+        stream.emit("Eighth message")
+        stream.emit("Ninth message")
+        stream.emit("Tenth message")
+        stream.emit("Eleventh message")
+        stream.emit("Twelfth message")
+
+        await asyncio.sleep(0.07)  # Wait for the flush task to complete
```
heyitsaamir: Instead of adding all these magic numbers, can we simply mock `call_later`? 

Basically you want to simulate how call_later will behave right? so mock that function, and have it trigger on some signal.

emit
emit
signal() <- this is your test to simulate the passing of some time. Not real time, but simulated time.
verify behavior 

Essentially, you know time is going to progress. You're not testing the behavior of call_later. So this is a reasonable way to test stuff like this.

packages/apps/tests/test_http_stream.py
```diff
@@ -63,37 +61,45 @@ def http_stream(self, mock_api_client, conversation_reference, mock_logger):
...
-        start_time = datetime.now()
-
         stream.emit("Test message")
-        await asyncio.sleep(0.6)  # Wait longer than 500ms timeout
-
-        assert mock_api_client.send_call_count > 0, "Should have sent at least one message"
-        assert any(t >= start_time + timedelta(milliseconds=450) for t in mock_api_client.send_times), (
-            "Should have waited approximately 500ms before sending"
-        )
+        await asyncio.sleep(0.07)  # Wait for the flush task to complete
```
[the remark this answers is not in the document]
MehakBindra: removing delays

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -126,6 +120,20 @@ def update(self, text: str) -> None:
...
+        async def _poll():
+            while not self._id or self._queue:
+                self._logger.debug("waiting for _id to be set or queue to be empty")
+                await asyncio.sleep(0.1)
+
+        try:
+            await asyncio.wait_for(_poll(), timeout=self._total_wait_timeout)
+            return True
+        except asyncio.TimeoutError:
+            return False
```
heyitsaamir: why do we need this?

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -216,13 +225,14 @@ async def _flush(self) -> None:
...
                 await self._send_activity(to_send)
 
-            # Signal if queue is now empty
-            if not self._queue:
-                self._queue_empty_event.set()
-
             # If more queued, schedule another flush
             if self._queue and not self._timeout:
-                self._timeout = Timeout(0.5, self._flush)
+                self._timeout = asyncio.get_running_loop().call_later(0.5, lambda: asyncio.create_task(self._flush()))
```
heyitsaamir: just using self.flush is ok too right? do we need the lambda?

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -104,18 +103,13 @@ def emit(self, activity: Union[MessageActivityInput, TypingActivityInput, str])
...
         if isinstance(activity, str):
             activity = MessageActivityInput(text=activity, type="message")
         self._queue.append(activity)
 
-        # Clear the queue empty event since we just added an item
-        self._queue_empty_event.clear()
-
-        self._timeout = Timeout(0.5, self._flush)
+        if not self._pending and not self._timeout:
+            self._pending = asyncio.create_task(self._flush())
```
heyitsaamir: We schedule the first task immediately to begin streaming.

heyitsaamir: Add comment on why no timeout is set

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -126,6 +120,20 @@ def update(self, text: str) -> None:
...
+
+        async def _poll():
+            while not self._id or self._queue:
+                self._logger.debug("waiting for _id to be set or queue to be empty")
+                await asyncio.sleep(0.1)
+
+        try:
+            await asyncio.wait_for(_poll(), timeout=self._total_wait_timeout)
+            return True
+        except asyncio.TimeoutError:
```
heyitsaamir: log when there's a timeout

MehakBindra: logged in close() function

PR #202, opened 2025-11-04
get type hint of class, not object

mjamroz:
Otherwise it stops working with python 3.14


lilyydu (approved):
Thanks for adding this!

PR #233, opened 2025-12-08
update react and react-dom to 19.2.1

lilyydu:
updated for samples/tab

link to issue:
[GHSA-9qr9-h5gf-34mp](https://github.com/vercel/next.js/security/advisories/GHSA-9qr9-h5gf-34mp)


packages/devtools/src/microsoft_teams/devtools/web/index.html
```diff
@@ -5,8 +5,8 @@
     <link rel="icon" type="image/svg+xml" href="/devtools/icon.png" />
```
heyitsaamir: Is there a file that needs to be removed?

lilyydu: I thought I removed  it but not sure why the diff isn't capturing it. let me db check

singhk97 (approved):

PR #241, opened 2026-01-05
Add support for Targeted Messages

ShanmathiMayuramKrithivasan:
Add support for Targeted Messages
This PR introduces support for sending targeted messages - messages delivered privately to a specific recipient within a conversation.

Key Updates:
Added an is_targeted boolean parameter to the send, update, reply and Delete APIs. When enabled, the message is sent privately to the Recipient.Id specified in the activity payload.

We append isTargetedActivity=true as a query parameter in API URLs when isTargeted is set, allowing backend services to correctly process these requests.

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/app_process.py
+++ b/packages/apps/src/microsoft_teams/apps/app_process.py
@@ -131,9 +131,12 @@ class ActivityProcessor:
             message: str | ActivityParams | AdaptiveCard,
             conversation_ref: Optional[ConversationReference] = None,
             *,
+            is_targeted: bool = False,
             targeted_recipient_id: Optional[str] = None,
         ) -> SentActivity:
-            res = await send(message, conversation_ref, targeted_recipient_id=targeted_recipient_id)
+            res = await send(
+                message, conversation_ref, is_targeted=is_targeted, targeted_recipient_id=targeted_recipient_id
+            )
 
             if not self.event_manager:
                 raise ValueError("EventManager was not initialized properly")


packages/api/src/microsoft_teams/api/clients/conversation/activity.py
```diff
@@ -58,55 +62,77 @@ async def create(self, conversation_id: str, activity: ActivityParams) -> SentAc
...
 
         Args:
             conversation_id: The ID of the conversation
             activity_id: The ID of the activity to delete
+            is_targeted: When True, deletes a targeted message privately
         """
-        await self.http.delete(f"{self.service_url}/v3/conversations/{conversation_id}/activities/{activity_id}")
+        url = f"{self.service_url}/v3/conversations/{conversation_id}/activities/{activity_id}"
+        if is_targeted:
+            url += "?isTargetedActivity=true"
```
heyitsaamir: do you need to have "isTargeted" for delete messages?

ShanmathiMayuramKrithivasan: Yes based on service side design, we need to pass isTargetedActivity for update and delete.

packages/apps/src/microsoft_teams/apps/app.py
```diff
@@ -259,8 +259,17 @@ async def on_http_stopped() -> None:
             self._events.emit("error", ErrorEvent(error, context={"method": "stop"}))
             raise
 
-    async def send(self, conversation_id: str, activity: str | ActivityParams | AdaptiveCard):
-        """Send an activity proactively."""
+    async def send(
+        self, conversation_id: str, activity: str | ActivityParams | AdaptiveCard, is_targeted: bool = False
```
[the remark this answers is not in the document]
ShanmathiMayuramKrithivasan: With this design, we are staying consistent with the .NET SDK design and the other API client methods by keeping it as a boolean parameter.

packages/apps/src/microsoft_teams/apps/app.py
```diff
@@ -279,7 +288,11 @@ async def send(self, conversation_id: str, activity: str | ActivityParams | Adap
         else:
             activity = activity
 
-        return await self.http.send(activity, conversation_ref)
+        # Validate recipient for targeted messages
+        if is_targeted and not activity.recipient:
+            raise ValueError("activity.recipient is required for targeted messages")
```
lilyydu: is there any discussion in the future that there could be multiple recipients for targeted msgs?

packages/apps/src/microsoft_teams/apps/contexts/function_context.py
```diff
@@ -80,7 +86,11 @@ async def send(self, activity: str | ActivityParams | AdaptiveCard) -> Optional[
         else:
             activity = activity
 
-        return await self.http.send(activity, conversation_ref)
+        # Validate recipient for targeted messages
+        if is_targeted and not activity.recipient:
+            raise ValueError("activity.recipient is required for targeted messages")
```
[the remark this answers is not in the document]
lilyydu: what happens if the intended recipient is not in the conversation?

packages/apps/src/microsoft_teams/apps/app.py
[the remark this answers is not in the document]
heyitsaamir: We should fix it in C# as well. This design isn't ergonomic. It's not clear for the user that if `activity` was string or adaptive card, that setting `is_targeted` will throw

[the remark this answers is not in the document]
heyitsaamir: User should be able to do `app.send(conversation_id, "hello", targeted_recipient_id=<user_id>)`

packages/devtools/src/microsoft_teams/devtools/devtools_plugin.py
```diff
@@ -225,8 +225,10 @@ async def on_activity_response(self, event: PluginActivityResponseEvent):
             promise.set_result(event.response)
             del self.pending[event.activity.id]
 
-    async def send(self, activity: ActivityParams, ref: ConversationReference) -> SentActivity:
-        return await self.http.send(activity, ref)
+    async def send(
+        self, activity: ActivityParams, ref: ConversationReference, is_targeted: bool = False
```
heyitsaamir: was this tested on devtools? (just making sure it doesn't crash it)

packages/api/src/microsoft_teams/api/clients/conversation/activity.py
```diff
@@ -50,12 +50,14 @@ async def create(
...
-        params = {"isTargetedActivity": "true"} if is_targeted else None
+        params: dict[str, str] = {}
+        if is_targeted:
+            params["isTargetedActivity"] = "true"
 
         response = await self.http.post(
             url,
             json=activity.model_dump(by_alias=True, exclude_none=True),
-            params=params,
+            params=params or None,
```
lilyydu: nit: don't think this can be `None` since it was declared previously

packages/apps/src/microsoft_teams/apps/contexts/function_context.py
```diff
@@ -80,7 +86,12 @@ async def send(self, activity: str | ActivityParams | AdaptiveCard) -> Optional[
         else:
             activity = activity
 
-        return await self.http.send(activity, conversation_ref)
+        # Handle targeted messages
+        is_targeted = targeted_recipient_id is not None
+        if is_targeted:
+            activity.recipient = Account(id=targeted_recipient_id)
```
heyitsaamir: if `activity.recipient` is already provided, do a verification that it's the same as the provided `targeted_recipient_id` and throw otherwise.

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
```diff
@@ -171,13 +172,15 @@ async def send(
         self,
         message: str | ActivityParams | AdaptiveCard,
         conversation_ref: Optional[ConversationReference] = None,
+        targeted_recipient_id: Optional[str] = None,
```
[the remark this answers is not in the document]
ShanmathiMayuramKrithivasan: Agreed, the common case should be simple. Added is_targeted with targeted_recipient_id - if both are set, targeted_recipient_id takes precedence.

packages/api/src/microsoft_teams/api/clients/conversation/activity.py
heyitsaamir: This is being discussed

packages/api/src/microsoft_teams/api/clients/conversation/activity.py
```diff
@@ -58,55 +62,77 @@ async def create(self, conversation_id: str, activity: ActivityParams) -> SentAc
...
         Args:
             conversation_id: The ID of the conversation
             activity_id: The ID of the activity to update
             activity: The updated activity data
+            is_targeted: When True, sends the message privately to the recipient specified in activity.recipient
 
         Returns:
             The updated activity
         """
+        url = f"{self.service_url}/v3/conversations/{conversation_id}/activities/{activity_id}"
```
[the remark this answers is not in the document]
heyitsaamir: This is being discussed

PR #270, opened 2026-02-04
remove custom logger, use standard logging library

lilyydu:
Previously, we had a custom logger available for customers. However, this didn't integrate very well with the general industry - other libraries and frameworks use python's `logging` library (e.g., fastapi, boto3, etc), and most developers are familiar with it. In addition, using the standard library is better for reliability, and requires less maintenance.

- updated to use module level logging throughout the packages (removed from constructors, dependency injections)
- removed the `ConsoleLogger`, `LoggerDependencyOption`, and `EventEmitterOptions` (it only has a logger property)
- kept the `ConsoleFormatter` and `ConsoleFilter` as an option for developers to use
- added the `NullHandler` so that no errors/warnings will show up when logger has not yet been configured (added to every package init file)
- updated common/README to explain how to setup and customize the logger
- removed mock loggers from tests
- regenerated the handlers now that logger is no longer a property

example w/ standard logger, using our custom formatter:
image

Sharing notes/observations. I looked at the most popular python libraries, as well as the most popular multi-package sdks.

**Pattern 1 - Single top-level logger (FastAPI, httpx, stripe)**
Simple, all log output is under one namespace, but all of the output is in one undifferentiated stream. No granularity.
**Pattern 2 - Per module logger (OpenAI, Anthropic, urllib4, botocore, azure sdks, google cloud sdks)**
Fine-grained, names expose the internal implementation details
**Pattern 3 - Central log.py file, with named subsystems (aiohttp)**
Stable, intuitive because users filter by what the system is doing since all logger definitions live in one spot.
Very much overkill, requires more coordination, rare design if subsystems don't need such independent filtering.

If users don't want to see any logs, thats what `NullHandler` is for - logs are silent and python warnings are suppressed.

**When users want to see all the logs:**
logging.getLogger("microsoft_teams").setLevel(logging.DEBUG)

**When users want to narrow it down:**
logging.getLogger("microsoft_teams.ai").setLevel(logging.DEBUG)

### DIFF
--- a/packages/common/tests/test_backward_compat.py
+++ b/packages/common/tests/test_backward_compat.py
@@ -69,6 +69,8 @@ def test_deprecation_warning_message_content():
     with warnings.catch_warnings(record=True) as w:
         warnings.simplefilter("always")
 
+        from microsoft.teams.common import Client  # noqa: F401
+
         assert len(w) >= 1, f"Expected deprecation warning, got {len(w)} warnings"
         warning_msg = str(w[0].message)
 


packages/apps/src/microsoft_teams/apps/options.py
```diff
@@ -38,6 +38,11 @@ class AppOptions(TypedDict, total=False):
 
     # Infrastructure
     logger: Optional[Logger]
+    """
+    DEPRECATED:
+    This method of setting the logger is deprecated and will be removed in version 2.0.0 GA."
```
corinagum:
```suggestion
    This method of setting the logger is deprecated and will be removed in version 2.0.0 GA.
```
Just removing stray quotation

corinagum (commented):
two questions
- does `app_oauth.py` need to be migrated to `logging.getLogger`?
- do we plan to update the examples as well?

lilyydu:
@heyitsaamir 
I think per-module loggers makes the most sense given our multiple packages. It allows more granularity (silencing a certain module's logs for debugging), flexibility, and shows hierarchy for free. OpenAI/Anthropic are most closely similiar to us and they both do the same.

---

Sharing notes/observations. I looked at the most popular python libraries, as well as the most popular multi-package sdks.

**Pattern 1 - Single top-level logger (FastAPI, httpx, stripe)**
- Simple, all log output is under one namespace, but all of the output is in one undifferentiated stream. No granularity.

**Pattern 2 - Per module logger (OpenAI, Anthropic, urllib4, botocore, azure sdks, google cloud sdks)**
- Fine-grained, names expose the internal implementation details

**Pattern 3 - Central `log.py` file, with named subsystems (aiohttp)**
- Stable, intuitive because users filter by what the system is doing since all logger definitions live in one spot.
- Very much overkill, requires more coordination, rare design if subsystems don't need such independent filtering.

---

**If users don't want to see any logs,** thats what `NullHandler` is for - logs are silent and python warnings are suppressed.

**When users want to see all the logs:**
`logging.getLogger("microsoft_teams").setLevel(logging.DEBUG)`

**When users want to narrow it down:**
logging.getLogger("microsoft_teams.ai").setLevel(logging.DEBUG)

PR #273, opened 2026-02-09
fix: get_participant in meetings client to include tenantId

lilyydu:
original issue in https://github.com/microsoft/teams.net/issues/276
similiar fix to https://github.com/microsoft/teams.net/pull/277

locally tested:
image

### DIFF
--- a/packages/api/src/microsoft_teams/api/clients/meeting/client.py
+++ b/packages/api/src/microsoft_teams/api/clients/meeting/client.py
@@ -45,16 +45,19 @@ class MeetingClient(BaseClient):
         response = await self.http.get(f"{self.service_url}/v1/meetings/{id}")
         return MeetingInfo.model_validate(response.json())
 
-    async def get_participant(self, meeting_id: str, id: str) -> MeetingParticipant:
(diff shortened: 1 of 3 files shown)


MehakBindra (approved):

packages/api/src/microsoft_teams/api/clients/meeting/client.py
```diff
@@ -45,16 +45,19 @@ async def get_by_id(self, id: str) -> MeetingInfo:
...
             meeting_id: The meeting ID.
             id: The participant ID.
+            tenant_id: The tenant ID.
 
         Returns:
             The meeting participant information.
         """
-        response = await self.http.get(f"{self.service_url}/v1/meetings/{meeting_id}/participants/{id}")
+
+        url = f"{self.service_url}/v1/meetings/{meeting_id}/participants/{id}?tenantId={tenant_id}"
```
corinagum: In TS, I went ahead and updated this and the url above with URI encoding to more closely match C# & be more technically correct even though it's not strictly necessary. Do you want to do that here?

PR #291, opened 2026-02-24
move nbgv-python install into release group

lilyydu:
moved `nbgv-python` into a separate "release" dependency-group, so that `uv sync` doesnt throw an error when developer doesnt have it installed

error that I was getting:
```
TypeError: NbgvCommandError.__init__() missing 3 required positional arguments: 'returncode', 'stdout', and 'stderr'
      hint: This usually indicates a problem with the package or the build environment.
```
no impact on existing release pipeline/samples:
image

to re-install:
run `uv sync --group release`

### DIFF
--- a/.azdo/publish.yml
+++ b/.azdo/publish.yml
@@ -18,7 +18,9 @@ parameters:
   - Public
 
 variables:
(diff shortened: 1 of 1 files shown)


corinagum (approved):
I also tested locally - LGTM! I think having a comment would be good in readme, but non-blocking

PR #313, opened 2026-03-14
Route BotBuilder auth through HttpServer.handle_request

heyitsaamir:
## Summary
- Make `HttpServer.handle_request` public so plugins can route through SDK-level JWT validation
- Change DI from `HttpServerAdapter` to `HttpServer` — plugins now inject `HttpServer` and access the adapter via `.adapter`
- BotBuilderPlugin runs CloudAdapter auth first, then calls `http_server.handle_request()` for SDK auth + Teams pipeline (previously bypassed `TokenValidator` entirely)
- Remove pass-through methods (`register_route`, `serve_static`, `start`, `stop`) from `HttpServer` — callers go through `server.adapter` directly

## Test plan
- [x] `uv run pyright` — no new type errors
- [x] `poe test` — all apps, botbuilder, mcpplugin tests pass (153 tests)

### DIFF
--- a/packages/botbuilder/src/microsoft_teams/botbuilder/botbuilder_plugin.py
+++ b/packages/botbuilder/src/microsoft_teams/botbuilder/botbuilder_plugin.py
@@ -6,18 +6,16 @@ Licensed under the MIT License.
 import importlib.metadata
 from logging import Logger
 from types import SimpleNamespace
-from typing import Annotated, Any, Callable, Dict, Optional, TypedDict, Unpack, cast
+from typing import Annotated, Optional, TypedDict, Unpack, cast
 
-from microsoft_teams.api import Credentials, InvokeResponse
+from microsoft_teams.api import Credentials
 from microsoft_teams.apps import (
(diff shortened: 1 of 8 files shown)


packages/botbuilder/src/microsoft_teams/botbuilder/botbuilder_plugin.py
```diff
@@ -127,39 +123,8 @@ async def logic(turn_context: TurnContext) -> None:
...
-                resp_dict = cast(Dict[str, Any], result)
-
-            if "status" in resp_dict:
-                status_code = int(resp_dict.get("status", 200))
-            if "body" in resp_dict:
-                resp_body = resp_dict.get("body")
-
-            return HttpResponse(status=status_code, body=resp_body)
+            # Route through HttpServer for SDK auth + Teams pipeline
+            return await self.http_server.handle_request(request)
```
lilyydu: this means we'll validate each token twice right? once in `CloudAdapter.process_activity` and once again here

PR #337, opened 2026-03-31
Implement stream cancelation

### LINKED ISSUE #332: [Bug]: HttpStream doesn't handle canceling of stream
### Bug Description

It would be nice if this could be supported, I have implemented it myself now in a wrapper for `HttpStream`. If anyone is interested in it:

```
class CustomHttpStream(HttpStream):
    def __init__(self, client: ApiClient, ref: ConversationReference):
        super().__init__(client, ref)
        self._is_canceled = False

    def emit(self, activity: Union[MessageActivityInput, TypingActivityInput, str]) -> None:
        if self.is_canceled():
            logger.warning("Not emitting activity, stream is canceled.")
            return None
        return super().emit(activity)

    async def _send(self, to_send: Union[TypingActivityInput, MessageActivityInput]) -> SentActivity:
        try:
            res = await super()._send(to_send)
            return res
        except HTTPStatusError as e:
            if e.response.status_code == 403:
                # Set canceled and return a `CancelledError` so that the
                # request won't be retried.
                self._is_canceled = True
                logger.warning("Teams channel stopped the stream.")
                raise asyncio.CancelledError("Teams channel stopped the stream.")
            raise e
        except Exception as e:
            raise e

    def is_canceled(self) -> bool:
        return self._is_canceled
```

### Steps to Reproduce

1. Start a stream
2. Send messages
3. Cancel from Teams
4. `HttpStream` tries to continue sending updates in a retry, no way to see that it was canceled

### Expected Behavior

`HttpStream` should stop attempt sending updates, and it should be possible to see that it was canceled.

### Actual Behavior

`HttpStream` tries to continue sending updates in a retry, there is no way to see that it was canceled

### SDK Version

20.0.0a20

### Python Version

3.14.2

### Additional Context

_No response_

jerbob92:
Fixes #332

### DIFF
--- a/examples/ai-test/src/main.py
+++ b/examples/ai-test/src/main.py
@@ -4,6 +4,7 @@ Licensed under the MIT License.
 """
 
 import asyncio
+import logging
 import re
 from os import getenv
 
@@ -29,6 +30,8 @@ from microsoft_teams.devtools import DevToolsPlugin
 from microsoft_teams.openai import OpenAICompletionsAIModel, OpenAIResponsesAIModel
 
 load_dotenv(find_dotenv(usecwd=True))
+
+logging.basicConfig(level=getenv("LOG_LEVEL", "WARNING").upper())
 
 
 def get_required_env(key: str) -> str:
--- a/packages/apps/src/microsoft_teams/apps/app_process.py
+++ b/packages/apps/src/microsoft_teams/apps/app_process.py
@@ -27,7 +27,7 @@ if TYPE_CHECKING:
 
 from .activity_sender import ActivitySender
 from .events import ActivityEvent, ActivityResponseEvent, ActivitySentEvent, ErrorEvent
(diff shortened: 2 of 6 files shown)


heyitsaamir:
> > * Emit should be a no-op if the stream is cancelled.
> 
> What do you mean? It already is.
> 
> > * We potentially need a way to short-circuit the AI generation if emit is short-circuiting. One approach is to throw on emit if stream is cancelled. But then we'd break all existing handlers. Another approach is for the handlers to have access to this signal, and choose to short-circuit themselves.
> 
> I'd like the second approach more. Perhaps something using `yield`? (I'm not really a Python dev so not sure)
> 
> Edit: Nvm, I see there is some `EventEmitter` system in place already, better to use that.

Oops you're right. Ignore 1st.

@jerbob92 why do you prefer second? I think your `prompt.send` throwing if the stream gets cancelled is ok.

PR #354, opened 2026-04-03
Add sovereign cloud support (GCCH, DoD, China)

corinagum:
## Summary
- Introduces `CloudEnvironment` frozen dataclass with predefined presets (`PUBLIC`, `US_GOV`, `US_GOV_DOD`, `CHINA`) bundling all cloud-specific service endpoints
- Threads cloud environment through `App`, `TokenManager`, `BotTokenClient`, `TokenValidator`, `ApiClient`, and `ApiClientSettings`
- Supports `CLOUD` environment variable and programmatic `AppOptions.cloud` configuration
- Adds `graph_scope` to `CloudEnvironment` for cloud-aware Microsoft Graph token acquisition
- Simplifies `merge_api_client_settings` with `cloud` default to `PUBLIC`, removes `DEFAULT_API_CLIENT_SETTINGS`
- Fixes cloud propagation through `ApiClient` -> `BotClient` -> `BotTokenClient`

### Note
`graph_base_url` (Graph API endpoint per cloud) is intentionally deferred. This PR focuses on auth/token acquisition. Graph API routing is a separate concern.

### Sources
- Bot Framework GCCH/DoD endpoints: https://learn.microsoft.com/en-us/azure/bot-service/how-to-deploy-gov-cloud-high
- Bot Framework China endpoints: https://learn.microsoft.com/en-us/azure/bot-service/how-to-deploy-china-cloud
- Graph national cloud deployments: https://learn.microsoft.com/en-us/graph/deployments

## Test plan
- [x] `pytest packages/api/tests/` -- 168 tests pass
- [x] `pytest packages/apps/tests/` -- 389 tests pass (1 pre-existing failure unrelated)
- [x] E2E: Echo bot with `CLOUD=USGov` against real GCCH tenant -- message received, echo reply sent
- [x] Copilot review feedback addressed

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/api/src/microsoft_teams/api/auth/__init__.py
+++ b/packages/api/src/microsoft_teams/api/auth/__init__.py
@@ -4,6 +4,15 @@ Licensed under the MIT License.
 """
 
 from .caller import CallerIds, CallerType
+from .cloud_environment import (
+    CHINA,
+    PUBLIC,
+    US_GOV,
+    US_GOV_DOD,
+    CloudEnvironment,
+    from_name,
+    with_overrides,
+)
 from .credentials import (
     ClientCredentials,
     Credentials,
@@ -17,12 +26,19 @@ from .token import TokenProtocol
 __all__ = [
     "CallerIds",
     "CallerType",
+    "CHINA",
+    "CloudEnvironment",
     "ClientCredentials",
     "Credentials",
     "FederatedIdentityCredentials",
+    "from_name",
     "ManagedIdentityCredentials",
+    "PUBLIC",
     "TokenCredentials",
     "TokenProtocol",
     "JsonWebToken",
     "JsonWebTokenPayload",
+    "US_GOV",
+    "US_GOV_DOD",
+    "with_overrides",
 ]
--- a/packages/api/src/microsoft_teams/api/auth/__init__.py
+++ b/packages/api/src/microsoft_teams/api/auth/__init__.py
@@ -5,14 +5,11 @@ Licensed under the MIT License.
 
 from .caller import CallerIds, CallerType
 from .cloud_environment import (
-    CHINA,
-    PUBLIC,
-    US_GOV,
-    US_GOV_DOD,
     CloudEnvironment,
     from_name,
     with_overrides,
 )
+from .cloud_environment import from_name as config_from_cloud_name
 from .credentials import (
     ClientCredentials,
     Credentials,
@@ -26,19 +23,16 @@ from .token import TokenProtocol
 __all__ = [
     "CallerIds",
     "CallerType",
-    "CHINA",
     "CloudEnvironment",
     "ClientCredentials",
+    "config_from_cloud_name",
     "Credentials",
     "FederatedIdentityCredentials",
     "from_name",
     "ManagedIdentityCredentials",
-    "PUBLIC",
     "TokenCredentials",
     "TokenProtocol",
     "JsonWebToken",
     "JsonWebTokenPayload",
-    "US_GOV",
-    "US_GOV_DOD",
     "with_overrides",
 ]
--- a/packages/api/src/microsoft_teams/api/auth/__init__.py
+++ b/packages/api/src/microsoft_teams/api/auth/__init__.py
@@ -6,7 +6,6 @@ Licensed under the MIT License.
 from .caller import CallerIds, CallerType
 from .cloud_environment import (
     CloudEnvironment,
-    from_name,
     with_overrides,
 )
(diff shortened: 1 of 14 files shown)


heyitsaamir (commented):
Hm, it looks like cloud simply overrides a number of application-level settings. Can we localize its spread to just `app`, and maybe the token-validator?? Basically if CLOUD is set, then we set many of those values up and log it?

rido-min (approved):

packages/api/src/microsoft_teams/api/auth/__init__.py
```diff
@@ -4,6 +4,15 @@
 """
 
 from .caller import CallerIds, CallerType
+from .cloud_environment import (
```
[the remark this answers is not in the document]
corinagum: Hmm when you say weird what do you mean? Unknown usage? I think out of context US_GOV being non-explanatory is valid, but the type annotation `cloud: CloudEnvironment` makes the context clear at point of use. If we do do a rename, "CLOUD_ENV_CONFIG_US_GOV" is really long. There are some options:

1. Don't export from top-level: remove from `auth/__init__.py` & `api/__init__.py`. Users import from the module directly: `from microsoft_teams.api.auth.cloud_environment import US_GOV`. The path makes the context clear.
2. Rename - I would prefer something shorter like `CLOUD_US_GOV`. 
3. Keep as-is: type annotation provides context.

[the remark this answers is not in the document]
corinagum: Sounds good, done.

packages/api/src/microsoft_teams/api/auth/__init__.py
```diff
@@ -4,6 +4,12 @@
...
+    CloudEnvironment,
+    from_name,
```
heyitsaamir: NIT: remove this import because we're importing it aliased below.

packages/api/src/microsoft_teams/api/auth/__init__.py
```diff
@@ -17,12 +23,16 @@
 __all__ = [
     "CallerIds",
     "CallerType",
+    "CloudEnvironment",
     "ClientCredentials",
+    "config_from_cloud_name",
     "Credentials",
     "FederatedIdentityCredentials",
+    "from_name",
```
heyitsaamir: Remove

packages/api/src/microsoft_teams/api/clients/api_client_settings.py
```diff
@@ -7,6 +7,10 @@
 from dataclasses import dataclass
 from typing import Optional
 
+from ..auth.cloud_environment import PUBLIC, CloudEnvironment
+
+DEFAULT_OAUTH_URL = "https://token.botframework.com"
```
heyitsaamir: @rajan-chari do you know if this needs to be different too?

packages/api/src/microsoft_teams/api/auth/cloud_environment.py
```diff
@@ -0,0 +1,119 @@
...
+
+US_GOV = CloudEnvironment(
+    login_endpoint="https://login.microsoftonline.us",
+    login_tenant="MicrosoftServices.onmicrosoft.us",
+    bot_scope="https://api.botframework.us/.default",
+    token_service_url="https://tokengcch.botframework.azure.us",
+    openid_metadata_url="https://login.botframework.azure.us/v1/.well-known/openidconfiguration",
+    token_issuer="https://api.botframework.us",
+    channel_service="https://botframework.azure.us",
+    oauth_redirect_url="https://tokengcch.botframework.azure.us/.auth/web/redirect",
```
heyitsaamir: Is this used anywhere? (oauth_redirect_url)

heyitsaamir (approved):
Looks good. Some minor nits. Also, looks like not all the fields are actually used. Is that intentional?

heyitsaamir (approved):

packages/api/src/microsoft_teams/api/auth/cloud_environment.py
corinagum: Removed as discussed

packages/api/src/microsoft_teams/api/clients/api_client_settings.py
corinagum: This is sentinel pattern that could be tightened up with these soveriegn changes. It's only checked for comparison to if the customer changes the value. Can just update for whether it is set at all to eliminate it.

lilyydu (approved):

PR #360, opened 2026-04-05
Reject unauthenticated when no credentials

heyitsaamir:
If a server is set up without any credentials, it currently will not be able to _send_ messages, but it _will_ currently accept all incoming requests.
This scenario is rare, but possible, and in this PR we reject any incoming requests if creds are not setup (and if `skip_auth` isn't explicitly set to `True`).

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/http/http_server.py
+++ b/packages/apps/src/microsoft_teams/apps/http/http_server.py
@@ -138,12 +138,9 @@ class HttpServer:
                     ),
                 )
             elif not self._token_validator:
-                # No credentials configured — reject the request
-                logger.error(
-                    "No credentials configured. Configure client authentication "
-                    "to securely receive messages, or set skip_auth=True to allow "
-                    "unauthenticated requests."
-                )
+                # No credentials configured: reject the request. A startup
+                # warning was already emitted; logging per request would
+                # just add noise without new information.
                 return HttpResponse(status=401, body={"error": "Authentication not configured"})
             else:
                 if not authorization.startswith("Bearer "):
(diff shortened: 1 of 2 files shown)


packages/apps/src/microsoft_teams/apps/http/http_server.py
```diff
@@ -119,6 +111,28 @@ async def handle_request(self, request: HttpRequest) -> HttpResponse:
                         is_expired=lambda: False,
                     ),
                 )
+            elif not self._token_validator:
+                # No credentials configured — reject the request
+                logger.error(
+                    "No credentials configured. Configure client authentication "
+                    "to securely receive messages, or set skip_auth=True to allow "
+                    "unauthenticated requests."
```
corinagum: This will emit an error on every request, while the TS instead logs a warning on startup and then after that returns 401 silently. Should we align behavior?

corinagum: Aligned in `0ae065b` — stripped the per-request `logger.error`. Behavior now matches the rebased TS PR: one-time warning at startup, silent 401 per rejected request.

packages/apps/src/microsoft_teams/apps/http/http_server.py
```diff
@@ -119,6 +111,28 @@ async def handle_request(self, request: HttpRequest) -> HttpResponse:
...
+                    return HttpResponse(status=401, body={"error": "Unauthorized"})
+
+                raw_token = authorization.removeprefix("Bearer ")
+                service_url = cast(Optional[str], body.get("serviceUrl"))
+
+                try:
+                    await self._token_validator.validate_token(raw_token, service_url)
+                except Exception as e:
+                    logger.warning(f"JWT token validation failed: {e}")
+                    return HttpResponse(status=401, body={"error": "Unauthorized"})
```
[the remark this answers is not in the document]
corinagum: Addressed: the original warning call was already converted to %-style in `74d73f5`; the remaining f-string at the activity-received debug log is now also %-style in `0ae065b`.

lilyydu (approved):

PR #402, opened 2026-04-23
fix: route `app.get_app_graph` + `ctx.*_graph` to sovereign endpoints via `cloud.graph_scope`

corinagum:
- `App.get_app_graph()`, `ctx.app_graph`, and `ctx.user_graph` were routing all Microsoft Graph calls to the public-cloud endpoint (`https://graph.microsoft.com`) regardless of the sovereign cloud configured on the app. Sovereign customers (GCCH, DoD, China) would hit the wrong cloud. This PR derives the Graph base URL from `cloud.graph_scope` on `App` and plumbs it through every Graph client the framework constructs.
- Public cloud (default): `graph_scope = "https://graph.microsoft.com/.default"` → derived base URL = `https://graph.microsoft.com` = previous hardcoded default. **Zero behavior change.**
- Sovereign clouds: behavior changes from broken (silently public) to correct (per-cloud). This is the fix.

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/app_process.py
+++ b/packages/apps/src/microsoft_teams/apps/app_process.py
@@ -18,6 +18,7 @@ from microsoft_teams.api import (
     TokenProtocol,
     is_invoke_response,
 )
+from microsoft_teams.api.auth.cloud_environment import CloudEnvironment
 from microsoft_teams.api.clients.user.params import GetUserTokenParams
 from microsoft_teams.cards import AdaptiveCard
 from microsoft_teams.common import Client, ClientOptions, LocalStorage, Storage
@@ -49,6 +50,7 @@ class ActivityProcessor:
         token_manager: TokenManager,
         api_client_settings: Optional[ApiClientSettings],
         activity_sender: ActivitySender,
+        cloud: CloudEnvironment,
     ) -> None:
         self.router = router
         self.id = id
@@ -58,6 +60,7 @@ class ActivityProcessor:
         self.token_manager = token_manager
         self.api_client_settings = api_client_settings
         self.activity_sender = activity_sender
+        self.cloud = cloud
 
         # This will be set after the EventManager is initialized due to
         # a circular dependency
@@ -126,6 +129,7 @@ class ActivityProcessor:
             self.default_connection_name,
             activity_sender=self.activity_sender,
             app_token=lambda: self.token_manager.get_graph_token(tenant_id),
+            cloud=self.cloud,
         )
 
         send = activityCtx.send
--- a/packages/apps/src/microsoft_teams/apps/app_process.py
+++ b/packages/apps/src/microsoft_teams/apps/app_process.py
@@ -18,7 +18,7 @@ from microsoft_teams.api import (
     TokenProtocol,
     is_invoke_response,
 )
-from microsoft_teams.api.auth.cloud_environment import CloudEnvironment
+from microsoft_teams.api.auth.cloud_environment import PUBLIC, CloudEnvironment
 from microsoft_teams.api.clients.user.params import GetUserTokenParams
 from microsoft_teams.cards import AdaptiveCard
 from microsoft_teams.common import Client, ClientOptions, LocalStorage, Storage
@@ -50,7 +50,7 @@ class ActivityProcessor:
         token_manager: TokenManager,
         api_client_settings: Optional[ApiClientSettings],
         activity_sender: ActivitySender,
-        cloud: CloudEnvironment,
+        cloud: CloudEnvironment = PUBLIC,
     ) -> None:
         self.router = router
         self.id = id
--- a/packages/apps/src/microsoft_teams/apps/routing/activity_context.py
+++ b/packages/apps/src/microsoft_teams/apps/routing/activity_context.py
@@ -27,6 +27,7 @@ from microsoft_teams.api import (
     TokenExchangeState,
     TokenPostResource,
 )
+from microsoft_teams.api.auth.cloud_environment import CloudEnvironment
 from microsoft_teams.api.models.attachment.card_attachment import (
     OAuthCardAttachment,
     card_attachment,
@@ -83,6 +84,7 @@ class ActivityContext(Generic[T]):
         connection_name: str,
         activity_sender: ActivitySender,
         app_token: Token,
+        cloud: CloudEnvironment,
     ):
         self.activity = activity
         self.app_id = app_id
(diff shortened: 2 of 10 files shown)


packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -49,6 +50,7 @@ def __init__(
         token_manager: TokenManager,
         api_client_settings: Optional[ApiClientSettings],
         activity_sender: ActivitySender,
+        cloud: CloudEnvironment,
     ) -> None:
```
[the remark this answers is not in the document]
lilyydu: ^ should we set as optional w default?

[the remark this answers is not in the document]
corinagum: yep already done

packages/graph/src/microsoft_teams/graph/graph.py
```diff
@@ -6,19 +6,30 @@
...
+from msgraph.graph_request_adapter import GraphRequestAdapter
 from msgraph.graph_service_client import GraphServiceClient
 
 from .auth_provider import AuthProvider
 
 
-def get_graph_client(token: Optional[Token] = None) -> GraphServiceClient:
+def get_graph_client(
+    token: Optional[Token] = None,
+    base_url: Optional[str] = None,
```
heyitsaamir, line 22: let's make this a keyword arg instead of a positional arg

packages/graph/src/microsoft_teams/graph/graph.py
```diff
@@ -53,8 +55,17 @@ def get_token():
...
-        client = GraphServiceClient(credentials=credential)
-        return client
+
+        if base_url is None:
+            return GraphServiceClient(credentials=credential)
+
+        # Build a custom request adapter with the sovereign base URL.
+        # Normalize: strip any trailing slash on caller's input, then append "/v1.0/"
+        # to match the msgraph-sdk default shape ("https://graph.microsoft.com/v1.0/").
+        auth_provider = AzureIdentityAuthenticationProvider(credential)
```
heyitsaamir, line 66: This differs a bit from the code snippet in https://learn.microsoft.com/en-us/graph/sdks/national-clouds?tabs=python. Would you mind double checking if this is the preferred approach vs. what's documented here?

corinagum: MS Learn:
```
http_client = GraphClientFactory.create_with_default_middleware(host=NationalClouds.US_GOV)
adapter = GraphRequestAdapter(auth_provider, http_client)
graph_client = GraphServiceClient(request_adapter=adapter)
```

graph client:
```
auth_provider = AzureIdentityAuthenticationProvider(credential)
adapter = GraphRequestAdapter(auth_provider)
adapter.base_url = f"{base_url.rstrip('/')}/v1.0/"
return GraphServiceClient(request_adapter=adapter)
```

- current code does not use factory
- raw URL string instead of `NationalClouds`, which is only available for Python and PHP
- base url includes version 
- Supports any potential cloud (including hypothetical ones we do not know about)

Essentially, the differences are just tradeoff. More flexibility + closer matching to what TS and C# do, or making use of existing `msgraph_core` package.

heyitsaamir: Hmm, the former puts the responsibility of any cloud related stuff on the package right? Is the URL the only difference? or are there any other diffences between the various hosts for python's graph library?

There might be nuances that the python graph library has that we're not really aware of. I think using the recommended approach by them should be preferred in this case, unless you're sure no other nuances between clouds exist (and none will in the future either)

corinagum: There's no per-cloud branching anywhere in msgraph_core. The enum is referenced in exactly two places: factory URL construction (just string interpolation) and GraphTelemetryHandler.is_graph_url (a set-membership check). Neither does anything cloud-specific beyond validating it's a graph endpoint.

If msgraph ever adds per-cloud branching keyed on NationalClouds, we'd miss it. But adding such branching would be a breaking change to a (str, Enum) that all 4 raw-URL-based language SDKs (C#/Go/Java/TypeScript per MS Learn) would also be exposed to. Risk is low.

Changing to use the factory wouldn't support Bleu/EagleX/SCloud like the code currently does today, plus we lose the closer parity to TS and C#. Keeping as-is.

packages/apps/src/microsoft_teams/apps/utils/graph.py
```diff
@@ -3,15 +3,51 @@
...
+# Extracts scheme + host (+ optional port) from a URL-like scope such as
+# "https://graph.microsoft.us/.default" -> "https://graph.microsoft.us".
+_GRAPH_BASE_URL_RE = re.compile(r"^(https?://[^/]+)", re.IGNORECASE)
+
+
+def _derive_graph_base_url(cloud: Optional[CloudEnvironment]) -> Optional[str]:
+    """Derive the Graph API base URL from a cloud's graph_scope, or None if unavailable."""
+    if cloud is None:
+        return None
+    scope = (cloud.graph_scope or "").strip()
```
heyitsaamir: graph_scope is a str, not optional, so the `or` clause isn't needed

heyitsaamir: NIT: seems like it would be better to keep the url in the cloud contract and append `/.default` than the other way around

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
```diff
@@ -83,6 +84,7 @@ def __init__(
         connection_name: str,
         activity_sender: ActivitySender,
         app_token: Token,
+        cloud: CloudEnvironment = PUBLIC,
```
heyitsaamir, line 87: Ideally there should be 1 centralized place where the default cloud is being set. If it's not set, it should just be none. (like create_graph_client has cloud defaulted to none)

heyitsaamir (approved):
Won't block, but I would encourage a second think on

packages/apps/src/microsoft_teams/apps/utils/graph.py
corinagum: For parity this would mean changes to TS and C#; since sovereign changes are ongoing let's defer these changes for then.

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
corinagum: Same as above.

PR #404, opened 2026-04-24
Fix allowlist plumbing through through TokenValidator

corinagum:
`App(additional_allowed_domains=[...])` was silently ignored at the token-validation layer, rejecting user-approved service URLs (e.g. `canary.botapi.skype.com`) before they ever reached the allowlist-aware check in `HttpServer.handle_request`.

Fix threads the allowlist through `TokenValidator.__init__`, both factory methods (`for_service`, `for_entra`), and the `validate_token` call site. Updates `HttpServer.initialize` and `App.__init__` to forward the option.

## Why unit tests missed it

  All existing `additional_allowed_domains` tests used `skip_auth=True`, which bypasses the `TokenValidator` code path entirely. The helper function `is_allowed_service_url` had direct tests with 3 args, so the function itself was fine — just no test proved the code *calling* it in the auth-on path was also using 3 args. E2E validation was against default service url, which is in the built-in allowlist and doesn't need the knob. The intersection of "auth on" AND "non-default service URL" had zero coverage.

  ## Tests

  `test_token_validator.py`:
  - `test_validate_token_honors_additional_allowed_domains` — canary accepted when in allowlist
  - `test_validate_token_rejects_when_domain_not_in_allowlist` — baseline: canary rejected without the knob
  - `test_validate_token_wildcard_allows_arbitrary_domain` — `["*"]` disables the check at the token-validation layer
  - `test_for_service_stores_additional_allowed_domains` / `test_for_entra_stores_additional_allowed_domains` — factory plumbing

  `test_http_server.py`:
  - `test_initialize_forwards_allowlist_to_token_validator` — spies `TokenValidator.for_service` and asserts it's called *with* the allowlist (catches the exact regression
  class)

  All 67 tests pass.

  ## Test plan

  - [x] Unit: `uv run pytest packages/apps/tests/test_token_validator.py packages/apps/tests/test_http_server.py` — 67/67
  - [x] Lint/format: `uv run poe check` — clean
  - [x] Type check: `uv run pyright` — no new errors
  - [x] E2E on canary: verified activity from `https://canary.botapi.skype.com/amer/...` passes token validation and reaches `onMessage` (previously 401'd)
  - [x] E2E regression: default `smba.trafficmanager.net` still works

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/app.py
+++ b/packages/apps/src/microsoft_teams/apps/app.py
@@ -170,6 +170,7 @@ class App(ActivityHandlerMixin):
                 self.credentials.tenant_id,
                 application_id_uri=self.options.application_id_uri,
                 cloud=self.cloud,
+                additional_allowed_domains=self.options.additional_allowed_domains,
             )
 
     @property
--- a/packages/apps/src/microsoft_teams/apps/auth/token_validator.py
+++ b/packages/apps/src/microsoft_teams/apps/auth/token_validator.py
@@ -73,16 +73,23 @@ class TokenValidator:
     JWT token validator using PyJWKClient for simplified validation.
     """
 
-    def __init__(self, jwt_validation_options: JwtValidationOptions, cloud: Optional[CloudEnvironment] = None):
+    def __init__(
+        self,
+        jwt_validation_options: JwtValidationOptions,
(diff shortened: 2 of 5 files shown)


lilyydu (approved):

singhk97 (approved):
lgtm

PR #405, opened 2026-04-24
Add proactive message update example

singhk97:
## Summary
- Add Example 3 to the proactive-messaging sample showing how to send a message and then update it using the ConversationClient
- Demonstrates the send-then-update pattern for proactive messaging without a running server

## Test plan
- [x] Run `uv run src/main.py <conversation_id>` and verify all 3 examples work
- [x] Confirm the third message shows "Status: Pending..." then updates to "Status: Complete"


lilyydu (approved):
Nice! Thanks for updating!!

PR #435, opened 2026-05-22
feat: warn when bot starts without credentials

corinagum:
## Summary

- Add a startup warning in `HttpServer.initialize()` when no `CLIENT_ID` / `CLIENT_SECRET` / `TENANT_ID` is configured, so customers running anonymously know their bot accepts unauthenticated requests on `/api/messages`.
- 2 new unit tests in `test_http_server.py` covering: warning fires when no creds, warning does not fire when creds present.

## Why

Part of the DevTools deprecation rollout. The recommended local-testing path with Microsoft 365 Agents Playground is anonymous mode (no creds configured). The new warning makes that mode explicit at startup, so customers don't ship anonymous-mode bots to production by accident, and so the migration from DevTools to Playground produces a clear signal.

Behavior is unchanged. This is pure observability.

## Test plan

- [x] `pytest tests/test_http_server.py` (22/22 passing)
- [x] Confirmed warning text matches the wording used in teams.ts and teams.net for cross-SDK consistency
- [x] Verified at runtime by initializing `App()` without credentials. Output:
  ```
  [WARNING] microsoft_teams.apps.http.http_server: No credentials configured (CLIENT_ID / CLIENT_SECRET / TENANT_ID). Bot will accept unauthenticated requests on /api/messages.
  ```
  Note: Uvicorn's `dictConfig` replaces the default root-logger handler at server-startup, so consumers running through Uvicorn need their own logging config (or `logging.basicConfig(level=WARNING)`) to see this warning. Customers using the standard Agents templates already have logging configured.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/http/http_server.py
+++ b/packages/apps/src/microsoft_teams/apps/http/http_server.py
@@ -94,6 +94,12 @@ class HttpServer:
                 cloud=self._cloud,
             )
             logger.debug("JWT validation enabled for %s", self._messaging_endpoint)
+        elif not app_id:
+            logger.warning(
+                "No credentials configured (CLIENT_ID / CLIENT_SECRET / TENANT_ID). "
+                "Bot will accept unauthenticated requests on %s.",
+                self._messaging_endpoint,
+            )
 
         self._adapter.register_route("POST", self._messaging_endpoint, self.handle_request)
         self._initialized = True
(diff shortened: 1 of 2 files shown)


corinagum:
On the Copilot suggestion to add a separate warning when `skip_auth=True` is set with valid credentials: this is intentional. Cross-SDK design decision (same in teams.ts and teams.net) to warn only on missing credentials, not on the explicit `skip_auth` opt-in.

Rationale:
- `skip_auth=True` is an explicit user opt-in. Adding a warning would create noise for users who deliberately enable it (e.g., during local dev with Playground).
- The anonymous-mode warning is the safety net for the un-flagged case (no credentials configured), which is the easier mistake to make.

Marking as won't-fix for cross-SDK consistency.

lilyydu (approved):

PR #438, opened 2026-05-23
Allow unknown activity entities and missing channelData

### LINKED ISSUE #433: [Bug]: Activity validation fails on unknown entity types; strict tagged union rejects entire activity instead of ignoring unknown entities
### Bug Description

`microsoft-teams-apps` (and the underlying entity models in `microsoft-teams-api`) use a strict Pydantic discriminated union for activity `entities[]`. Any entity whose `type` discriminator is not in the SDK's known set causes `pydantic.ValidationError`, which the HTTP server returns as HTTP 500. **The entire activity is rejected**, not just the unknown entity — even when the rest of the payload is valid and the unknown entity is irrelevant to the handler.

Microsoft's Bot Framework service sends newer entity types (observed: `ClientCapabilities`) that the SDK doesn't recognize, so this breaks real-world bots. Impact is severe:

1. Repeated 5xx responses trip the **Teams channel's exponential backoff** and silently halt delivery to the bot. Recovery takes ~12–24h of consistent 2xx responses (in our case), or an Azure support ticket.
2. **Web Chat is affected** whenever Microsoft attaches an unknown entity to a `message` activity (not just `conversationUpdate`): the bot's `@app.on_message` handler never runs, so no reply is sent — even though Microsoft sees a 200 ack.
3. **Symptoms look like an Azure configuration problem** (tenant settings, manifest fields, AppId alignment, channel state), not an SDK bug, leading to long debugging sessions chasing the wrong layers.

Related closed issue: #237 (different specific field, same class of strict-validation rejection).

### Steps to Reproduce

⚠️ **Not reproducible via the DevTools simulator** (`microsoft-teams-devtools` plugin's `localhost:3979/devtools/` UI). DevTools sends synthetic activity payloads that don't include the problematic entity types. The bug only manifests against real Microsoft bot service traffic.

1. Create a minimal bot following `examples/echo` with `microsoft-teams-apps==2.0.11`.
2. Deploy to a publicly reachable HTTPS endpoint (Azure Container App, App Service, etc.) reachable by Microsoft's Bot Framework service.
3. Configure an Azure Bot resource pointing at the endpoint.
4. In the Azure portal, open the Azure Bot resource → **Test in Web Chat** → send any message.
5. Observe in container logs: `pydantic.ValidationError` and `HTTP 500` on activities whose `entities[]` include `ClientCapabilities`.

Example activity payload Microsoft is sending:
```json
{
  "type": "message",
  "entities": [
    {
      "type": "ClientCapabilities",
      "supportsListening": true,
      "supportsSplashScreen": true,
      "supportsTts": true
    }
  ]
}
```

### Expected Behavior

The SDK should **gracefully ignore unknown entity types** instead of rejecting the whole activity. Microsoft's Bot Framework activity schema is documented as forward-compatible: new fields and entity types are added over time without breaking older consumers, and SDKs are expected to ignore what they don't recognize.

For comparison, the older `botbuilder-python` SDK parses unknown entities loosely and does not reject the activity.

With this behaviour, a `message` activity containing one unknown entity and several known ones would still be dispatched to `@app.on_message` with the known entities intact.

### Actual Behavior

`ActivityTypeAdapter.validate_python` (in `microsoft_teams/apps/app_process.py:178`) fails the entire activity with `pydantic.ValidationError`. The HTTP server returns 500. The bot's handler never runs, no reply is sent.

Container log excerpt:

```
[ERROR] microsoft_teams.apps.http.http_server: N validation errors for tagged-union[...]
message.entities.0.ClientInfoEntity.type
  Input should be 'clientInfo' [type=literal_error, input_value='ClientCapabilities', input_type=str]
message.entities.0.MentionEntity.type
  Input should be 'mention' [type=literal_error, input_value='ClientCapabilities', input_type=str]
message.entities.0.MessageEntity.type
  Input should be 'https://schema.org/Message' [type=literal_error, input_value='ClientCapabilities', input_type=str]
... (repeats for AIMessageEntity, StreamInfoEntity, CitationEntity, SensitiveUsageEntity,
     ProductInfoEntity, QuotedReplyEntity, TargetedMessageInfoEntity)
INFO: POST /api/messages HTTP/1.1" 500 Internal Server Error
```

The SDK's known entity types (derived from `microsoft_teams/api/models/entity/*.py`):

```
clientInfo, mention, https://schema.org/Message, streaminfo,
ProductInfo, quotedReply, targetedMessageInfo
```

Microsoft sends an entity outside this set (`ClientCapabilities`) → Pydantic walks the discriminated union, finds no match, fails all branches, raises ValidationError → HTTP 500.

Additionally observed (different but related schema-strictness issue): `conversationUpdate` activities sometimes arrive without a `channelData` field, which the SDK marks as required, producing the same 500:

```
conversationUpdate.channelData
  Field required [type=missing, ...]
```

### SDK Version

microsoft-teams-apps: 2.0.11 microsoft-teams-api: 2.0.11

### Python Version

3.12

## Proposed fix — lenient unknown handling

Change the entity field's type from a strict tagged union to a structure that accepts unknown types, e.g.:

```python
entities: Optional[List[Union[KnownEntity, UnknownEntity]]] = None

class UnknownEntity(BaseModel):
    model_config = ConfigDict(extra="allow")
    type: str   # any string
```

Where `KnownEntity` is the existing tagged union of `ClientInfoEntity | MentionEntity | …`, and `UnknownEntity` is a fallback that preserves the raw payload without imposing a schema. The discriminator falls back to `UnknownEntity` when none of the known classes match.

This makes the SDK forward-compatible: future Microsoft additions don't break consumers; they just appear as `UnknownEntity` instances. Bot authors can opt in to handling them later. Ideally the same philosophy applies to `channelData` and other optional schema fields the SDK currently marks required.

## Workaround for anyone hitting this

FastAPI middleware that strips unknown entities from the request body before they reach `ActivityTypeAdapter`. Two phases: Phase 1 strips entities so the SDK can parse what's left; Phase 2 coerces residual 500s → 200s so Teams' channel backoff isn't tripped while a fix is in flight.

```python
_SDK_KNOWN_ENTITY_TYPES = {
    "clientInfo",
    "mention",
    "https://schema.org/Message",
    "streaminfo",
    "ProductInfo",
    "quotedReply",
    "targetedMessageInfo",
}

def _normalize_activity_payload(payload: dict) -> dict | None:
    modified = False
    entities = payload.get("entities")
    if isinstance(entities, list):
        kept = [e for e in entities if isinstance(e, dict) and e.get("type") in _SDK_KNOWN_ENTITY_TYPES]
        if len(kept) != len(entities):
            payload["entities"] = kept
            modified = True
    if payload.get("type") == "conversationUpdate" and "channelData" not in payload:
        payload["channelData"] = {}
        modified = True
    return payload if modified else None

@app.server.adapter.app.middleware("http")
async def normalize_and_safeguard_messages(request: Request, call_next):
    if request.url.path != "/api/messages":
        return await call_next(request)
    if request.method == "POST":
        try:
            body = await request.body()
            payload = json.loads(body)
            if isinstance(payload, dict):
                normalized = _normalize_activity_payload(payload)
                if normalized is not None:
                    new_body = json.dumps(normalized).encode()
                    async def receive():
                        return {"type": "http.request", "body": new_body, "more_body": False}
                    request._receive = receive
        except Exception:
            pass
    response = await call_next(request)
    if response.status_code == 500:
        return JSONResponse({"status": "acknowledged"}, status_code=200)
    return response
```

This is fragile — `_SDK_KNOWN_ENTITY_TYPES` has to stay in sync with the SDK's models on every upgrade. A proper SDK fix would let everyone delete it.

## Observed timeline / pattern

- We've seen Microsoft send `ClientCapabilities` initially only on `conversationUpdate` activities, later expanded to `message` activities too. Suggests Microsoft is rolling out new entity types incrementally; the SDK is consistently behind.
- After the bot started returning 5xx on these activities, Microsoft's Teams channel routing applied exponential backoff. Recovery (resumption of delivery) took roughly 12–24 hours of consistent 2xx responses, with an Azure support case open in parallel (unclear whether support intervened or backoff lifted naturally).

---

Happy to contribute a PR if there's interest. The two natural shapes are (1) add the missing entity types to the existing tagged union (narrow fix) or (2) make the entity union lenient about unknown types (forward-compatible fix). Let me know which direction fits the SDK's roadmap.

dcaayushd:
Fixes #433

## What changed
- Add a fallback UnknownEntity so new entity types do not break activity parsing
- Make conversationUpdate channelData optional to avoid 500s on missing data
- Add tests for unknown entities and missing channelData

## Why
Real Bot Framework traffic can include new entity types and sometimes skips channelData. These should not block message handling.

## Tests
- PYTHONPATH=packages/api/src:packages/common/src:packages/cards/src /Users/dcaayushd/Development/OpenSourceContribution/teams.py/venv/bin/python -m pytest packages/api/tests/unit/test_activity.py -k ActivityTypeAdapter
- PYTHONPATH=packages/api/src:packages/common/src:packages/cards/src /Users/dcaayushd/Development/OpenSourceContribution/teams.py/venv/bin/python -m pytest packages/api/tests/unit/test_activity.py


mjamroz:
> @mjamroz Could you please review this PR and approve the workflow when you have a chance? Thanks!

Hey @dcaayushd i believe it's some mistake, im not Microsoft staff neither related with teams.py

dcaayushd:
Sorry about the wrong tag! I'll find the right maintainer to review this.

dcaayushd:
@lilyydu  @corinagum  Could you please review this PR and approve the workflow when you have a chance? Thanks!

lilyydu:
> @lilyydu @corinagum Could you please review this PR and approve the workflow when you have a chance? Thanks!

Hey! Thanks for raising this fix - this requires a deeper discussion with our team. Will follow back when we reach a conclusion!

lilyydu:
Hi @dcaayushd,

Upon discussion, it is correct that Python should not be marking `channelData` as required. In both TS and .NET, this is a nullable property. 

And likewise for unknown entity types, this should not throw an error. 

Will leave review comments for fixes :)

packages/api/src/microsoft_teams/api/activities/conversation/conversation_update.py
```diff
@@ -49,7 +49,8 @@ class _ConversationUpdateBase(CustomBaseModel):
 
 
 class ConversationUpdateActivity(_ConversationUpdateBase, ActivityBase):
-    """Output model for received conversation update activities with required fields and read-only properties."""
-
-    channel_data: ConversationChannelData  # pyright: ignore [reportGeneralTypeIssues, reportIncompatibleVariableOverride]
```
lilyydu, line 54: Rather than moving completely, lets set this as `Optional[ConversationChannelData] = None`

packages/api/src/microsoft_teams/api/models/entity/unknown_entity.py
```diff
@@ -0,0 +1,18 @@
...
+
+from pydantic import ConfigDict
+
+from ..custom_base_model import CustomBaseModel
+
+
+class UnknownEntity(CustomBaseModel):
+    """Entity for unknown or forward-compatible types."""
+
+    model_config = ConfigDict(extra="allow")
```
lilyydu: `CustomBaseModel` already allows extra fields (https://github.com/microsoft/teams.py/blob/68771dbe25cc69a9621c4259da14f4c5575925ba/packages/api/src/microsoft_teams/api/models/custom_base_model.py#L46) so we don't need this part

packages/api/tests/unit/test_activity.py
```diff
@@ -7,16 +7,21 @@
 from datetime import datetime
 
 import pytest
+from microsoft_teams.api.activities import ActivityTypeAdapter
```
lilyydu: nit: combine imports from `microsoft_teams.api.activities`

packages/api/src/microsoft_teams/api/models/entity/unknown_entity.py
```diff
@@ -0,0 +1,18 @@
...
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from pydantic import ConfigDict
+
+from ..custom_base_model import CustomBaseModel
+
+
+class UnknownEntity(CustomBaseModel):
```
lilyydu: Hm.. another approach I am considering is instead of creating an UnknownEntity, we can create a base Entity, similiar to how C# does it. 

```python

 # entity_base.py  (new, or merged into entity.py)
 class EntityBase(CustomBaseModel):
     """Base entity. Unknown/forward-compatible types deserialize as this
     class."""
     type: str

 # mention_entity.py — inherit from EntityBase instead of CustomBaseModel
 class MentionEntity(EntityBase):
     type: Literal["mention"] = "mention"
     mentioned: Account
     text: Optional[str] = None

(same for ClientInfoEntity, MessageEntity, StreamInfoEntity, etc.)

 # entity.py — discriminated union with the base as fallback
 def _entity_discriminator(v: Any) -> str:
     t = v.get("type") if isinstance(v, dict) else getattr(v, "type", None)
     return t if t in _KNOWN else "_unknown"
 
 EntityUnion = Annotated[
     Union[
         Annotated[ClientInfoEntity, Tag("clientInfo")],
         Annotated[MentionEntity,    Tag("mention")],
         Annotated[MessageEntity,    Tag("https://schema.org/Message")],
         # ...
         Annotated[EntityBase,           Tag("_unknown")],   # ← base class IS the fallback
     ],
     Discriminator(_entity_discriminator),
 ]
```

I think this might be nicer b/c it futureproofs new entities. The tags and discriminator don't allow for any silent incorrect dispatching.  

cc @heyitsaamir on his thoughts

packages/api/src/microsoft_teams/api/activities/conversation/conversation_update.py
```diff
@@ -49,7 +49,11 @@ class _ConversationUpdateBase(CustomBaseModel):
...
 class ConversationUpdateActivity(_ConversationUpdateBase, ActivityBase):
-    """Output model for received conversation update activities with required fields and read-only properties."""
+    """Output model for received conversation update activities with read-only properties.
 
-    channel_data: ConversationChannelData  # pyright: ignore [reportGeneralTypeIssues, reportIncompatibleVariableOverride]
+    Note: channel_data may be absent for some conversationUpdate payloads
+    (e.g. Direct Line activities) and should be treated as optional.
+    """
+
+    channel_data: Optional[ConversationChannelData] = None  # pyright: ignore
```
heyitsaamir: Let's not ignore all of pyright. Can we be specific abotu the errors we're ignoring.

packages/api/src/microsoft_teams/api/models/entity/entity.py
```diff
@@ -16,15 +19,58 @@
...
+        Annotated[MentionEntity, Tag("mention")],
+        Annotated[MessageEntity, Tag("https://schema.org/Message")],
+        Annotated[AIMessageEntity, Tag("aiMessage")],
+        Annotated[StreamInfoEntity, Tag("streaminfo")],
+        Annotated[CitationEntity, Tag("citation")],
+        Annotated[SensitiveUsageEntity, Tag("sensitiveUsage")],
+        Annotated[ProductInfoEntity, Tag("ProductInfo")],
+        Annotated[QuotedReplyEntity, Tag("quotedReply")],
+        Annotated[TargetedMessageInfoEntity, Tag("targetedMessageInfo")],
+        Annotated[EntityBase, Tag("_unknown")],
```
heyitsaamir: I feel like this should be a dedicated `UnknownEntity(EntityBase)`. The main reason is that
If something comes in and is unknown, say: `{ "type": "foo" }`, it'll go through this logic and then come out `EntityBase`. Now, EntityBase is the superclass for all the other classes too, which makes it a bit more difficult to deduce that this isn't a supported entity type in the codebase. Also an `isInstance(<object>, EntityBase)` will be useless to determine if we want to check for unknown entity types.

heyitsaamir (commented):
Overall this looks good. I think for the unknown entity type, I'd prefer introducing a fallback `UnknownEntity` instead though. Curious if you had a different idea.

packages/api/src/microsoft_teams/api/models/entity/entity_base.py
```diff
@@ -0,0 +1,13 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from ..custom_base_model import CustomBaseModel
+
+
+class EntityBase(CustomBaseModel):
+    """Base entity for unknown and forward-compatible entity types."""
```
heyitsaamir: @copilot can you fix the comment here

heyitsaamir (approved):
Lgtm. Thanks for the contribution!

dcaayushd:
hey @lilyydu, this one's been approved by @heyitsaamir but needs a second review due to branch protection since he was the last to push. would you mind taking a look when you get a chance? thanks!

lilyydu (approved):

PR #439, opened 2026-05-26
Default targeted replies for targeted inbound messages

heyitsaamir:
## Summary
Make `ctx.send()` and `ctx.reply()` stay private by default when the inbound Teams message was targeted.

## Why
This matches the TS SDK behavior and saves app authors from manually wiring `with_recipient(..., is_targeted=True)` for the common private slash-command response path.

## Interesting bits
- Explicit public replies still work by passing a non-targeted recipient.
- Sends to a different conversation do not inherit targeted routing.
- Added typed `quotedReply` and `targetedMessageInfo` entities so the targeted reply metadata is modeled instead of shoved in as dict soup.
- Updated the targeted messages example with `send private` / `send public` commands.

## Tips for reviewers
Start with `packages/apps/src/microsoft_teams/apps/routing/activity_context.py`, then the new tests in `packages/apps/tests/test_activity_context.py`.

## Test plan

image

### DIFF
--- a/examples/targeted-messages/appPackage/manifest.json
+++ b/examples/targeted-messages/appPackage/manifest.json
@@ -1,0 +1,63 @@
+{
+    "$schema": "https://developer.microsoft.com/json-schemas/teams/vDevPreview/MicrosoftTeams.schema.json",
+    "version": "1.0.0",
+    "manifestVersion": "devPreview",
+    "id": "${{TEAMS_APP_ID}}",
+    "name": {"short": "targeted-messages", "full": "Targeted Messages Example"},
+    "developer": {
+        "name": "Microsoft",
+        "mpnId": "",
+        "websiteUrl": "https://microsoft.com",
+        "privacyUrl": "https://privacy.microsoft.com/privacystatement",
+        "termsOfUseUrl": "https://www.microsoft.com/legal/terms-of-use",
+    },
+    "description": {
+        "short": "Bot demonstrating targeted (ephemeral) messages",
+        "full": "A sample bot that demonstrates how to send targeted messages that are only visible to specific recipients in Microsoft Teams.",
+    },
+    "icons": {"outline": "outline.png", "color": "color.png"},
+    "accentColor": "#FFFFFF",
+    "bots": [
+        {
+            "botId": "${{BOT_ID}}",
+            "scopes": ["personal", "team", "groupChat"],
+            "isNotificationOnly": false,
+            "supportsCalling": false,
+            "supportsVideo": false,
+            "supportsFiles": false,
+            "supportsTargetedMessages": true,
+            "commandLists": [
+                {
+                    "scopes": ["team", "groupChat"],
+                    "triggers": ["slash"],
+                    "commands": [
+                        {"title": "test send", "description": "Send a targeted message visible only to you"},
+                        {
+                            "title": "test update",
+                            "description": "Send a targeted message then update it after 3 seconds",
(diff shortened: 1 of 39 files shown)


examples/targeted-messages/appPackage/manifest.json
```diff
@@ -1,55 +1,63 @@
...
+    "version": "1.0.0",
+    "manifestVersion": "devPreview",
+    "id": "${{TEAMS_APP_ID}}",
+    "name": {"short": "targeted-messages", "full": "Targeted Messages Example"},
+    "developer": {
+        "name": "Microsoft",
+        "mpnId": "",
+        "websiteUrl": "https://microsoft.com",
+        "privacyUrl": "https://privacy.microsoft.com/privacystatement",
+        "termsOfUseUrl": "https://www.microsoft.com/legal/terms-of-use",
```
corinagum: trailing `,` - does the manifest sideload without issue? If yes, feel free to disregard.

examples/targeted-messages/src/main.py
```diff
@@ -25,10 +25,33 @@ async def handle_message(ctx: ActivityContext[MessageActivity]):
...
-    if "test update" in text:
+    is_targeted_inbound = activity.recipient.is_targeted is True
+
+    print(
+        "[MESSAGE] Received:",
+        {"text": text, "is_targeted": is_targeted_inbound, "from": activity.from_.id},
+    )
+
+    if "send public" in text:
+        print("[send public]", {"is_targeted": is_targeted_inbound})
```
corinagum: I think in TS the console logs were removed, do we want to match that here, or keep? No strong opinion, just wanted to point it out.

corinagum (approved):

lilyydu (approved):
LGTM!

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
```diff
@@ -47,6 +49,7 @@
 
 T = TypeVar("T", bound=ActivityBase, contravariant=True)
 logger = logging.getLogger(__name__)
+_MessageActivityInputType = MessageActivityInput
```
lilyydu: curious, what does this do?

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
```diff
@@ -248,9 +292,9 @@ def _add_targeted_message_info_entity(self, activity_params: ActivityParams) ->
         and attaches the entity automatically so the developer doesn't need to.
         Skips if the developer already attached a targetedMessageInfo entity.
         """
-        if not self._is_incoming_targeted():
```
lilyydu, line 251: I think we had this as a bug before, but we should check conversation ref here to ensure its the same conversation (otherwise an explicit targeted send to a different conversation will still stamp the inbound activity's id as the targeted message id)

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
heyitsaamir: huh weird. we shouldn't need this.

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
heyitsaamir: Hmm good point, but it would only happen if the outbound is also targeted right? Maybe that's okay? Let's address this separately

PR #440, opened 2026-05-27
Update AI samples

MehakBindra:
Updating AI samples as follows : 

1. A2A - > Instead of the previous a2a + hitl huma operator flow, we now show hand off between 2 agents via a2a. 
2. Mcp server sample uses graph to get aad id of users in tenant via name before using notify, ask etc so we can ask agents to "message mehak to ... "
3. ai-mcp sample now has a local clarification tool instead of a welcome card tool, this is used whenever the agent requires human input for clarity.
4. introduced stream.clear to clear out existing streamed text, this is isefule for scenarios where the agent streams responses back to user but decides midway to send an adaptive card instead for eg.
5. Debugged conversation client to create conversation with user (exclude none = true)

https://github.com/user-attachments/assets/3bb2c8ae-9c4d-4297-824d-c2896601c8e3


examples/ai-mcp/src/main.py
```diff
@@ -0,0 +1,238 @@
...
+            type="application/vnd.microsoft.activity.message",
+            value="OK",
+        )
+    print(f"Running agent with clarification choice: {choice}")
+    conversation_id = ctx.activity.conversation.id
+    if conversation_id not in _sessions:
+        _sessions[conversation_id] = agent.create_session()
+
+    await _run_agent_and_reply(ctx, _sessions[conversation_id], choice)  # type: ignore[arg-type]
+
```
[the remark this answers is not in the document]
MehakBindra: works correctly

packages/apps/src/microsoft_teams/apps/plugins/streamer.py
```diff
@@ -81,6 +81,13 @@ def update(self, text: str) -> None:
         """
         ...
 
+    def clear_text(self) -> None:
+        """
+        Discard any text accumulated so far so the final activity is card-only.
+        Equivalent to ``stream.clearText()`` in the TypeScript SDK.
```
lilyydu: nit: don't think we need to mention TS SDK here, same in http_stream

examples/mcp-server/src/mcp_tools.py
```diff
@@ -75,15 +155,30 @@ async def request_approval(user_id: str, title: str, description: str) -> Approv
...
+    if entry is None:
         raise ValueError(f"No approval found with approval_id {approval_id}.")
-    return ApprovalResult(approval_id=approval_id, status=status)
+    if entry.status != "pending":
+        return ApprovalResult(approval_id=approval_id, status=entry.status)
+
+    try:
+        await asyncio.wait_for(entry.event.wait(), timeout=float(timeout_seconds))
+    except asyncio.TimeoutError:
+        pass
```
lilyydu: nit: logger here?

examples/mcp-server/src/mcp_tools.py
```diff
@@ -38,27 +83,62 @@ async def notify(user_id: str, message: str) -> NotifyResult:
...
+        actions=[
+            ExecuteAction(title="Send")
+            .with_data(SubmitData("ask_reply", {"request_id": request_id}))
+            .with_associated_inputs("auto"),
+        ],
+    )
+    try:
+        await app.send(conversation_id=conversation_id, activity=card)
+    except Exception:
+        pending_asks.pop(request_id, None)
```
lilyydu, line 111: nit: should we have logs throughout the exceptions in this class?

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -130,6 +130,21 @@ def update(self, text: str) -> None:
...
 
+    def clear_text(self) -> None:
+        """
+        Discard any text accumulated so far so the final activity is card-only.
+        Equivalent to ``stream.clearText()`` in the TypeScript SDK.
+
+        Clears the in-memory text buffer and drops any queued text-only
+        MessageActivityInput items that have not been flushed yet.
+        """
+        self._text = ""
```
lilyydu, line 147: do we need to hold the lock here, so its threadsafe? similiar to flush

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -130,6 +130,21 @@ def update(self, text: str) -> None:
...
+        self._queue = deque(
+            item for item in self._queue if not isinstance(item, MessageActivityInput) or item.attachments
```
lilyydu: I think we still need set `item.text = None` for the retained items, or clear text on all queued items regardless of attachment, cause in flush, we have `self._text += activity.text or ""`

packages/api/src/microsoft_teams/api/clients/conversation/client.py
```diff
@@ -144,6 +144,6 @@ async def create(self, params: CreateConversationParams) -> ConversationResource
         """
         response = await self.http.post(
             f"{self.service_url}/v3/conversations",
-            json=params.model_dump(by_alias=True),
+            json=params.model_dump(by_alias=True, exclude_none=True),
```
lilyydu, line 147: should this be False (to match PR desc?)

examples/a2a/.env.alice.template
```diff
@@ -0,0 +1,21 @@
...
+TENANT_ID=<Tenant-Id>
+
+# This bot listens on PORT; the A2A endpoint is mounted on the same server.
+PORT=3978
+
+# Shared Azure OpenAI for the LLM behind both bots.
+AZURE_OPENAI_ENDPOINT=https://<your-resource>.openai.azure.com
+AZURE_OPENAI_API_KEY=<your-api-key>
+AZURE_OPENAI_MODEL_DEPLOYMENT_NAME=<your-deployment-name>
+AZURE_OPENAI_API_VERSION=2024-10-21
```
lilyydu: I think this might be an extra var?

lilyydu: it looks like main file reads `BOT_APP_ID`, `BOT_APP_PASSWORD`, and agent calls _require_env("AZURE_OPENAI_MODEL"). - should we make these consistent

examples/a2a/src/a2a_client.py
```diff
@@ -0,0 +1,62 @@
...
+    @property
+    def cached_card(self) -> AgentCard | None:
+        return self._cached_card
+
+    async def get_peer_card(self) -> AgentCard:
+        """Fetch (and cache) the peer's AgentCard via its well-known endpoint."""
+        if self._cached_card:
+            return self._cached_card
+        if self._init_task is None:
+            self._init_task = asyncio.get_event_loop().create_task(self._resolve())
```
lilyydu: If `resolve()` raises an error, `_init_task` is set to a failed task persistently. I think we should wrap it in try/except:
```
 try:
     return await self._init_task
 except Exception:
     self._init_task = None
     logger.warning("some error msg")
     raise
```

lilyydu: I think asyncio.get_event_loop() is also deprecated in 3.10+; you should be able to use `asyncio.get_running_loop()` or `asyncio.create_task(self._resolve())`

examples/ai-mcp/src/main.py
```diff
@@ -0,0 +1,236 @@
...
+    """Generate 2 dynamic follow-up suggestions via a lightweight OpenAI call."""
+    try:
+        client = _get_openai_client()
+        model = getenv("AZURE_OPENAI_MODEL", "")
+        completion = await client.chat.completions.create(
+            model=model,
+            messages=[
+                {"role": "user", "content": last_user_text},
+                {"role": "assistant", "content": last_ai_text},
+                {"role": "system", "content": _FOLLOW_UPS_PROMPT},
```
lilyydu: nit: tmk system messages conventionally come first - could we move this to the start of the list?

examples/mcp-server/src/models.py
```diff
@@ -3,15 +3,19 @@
...
+from pydantic import BaseModel, ConfigDict, Field
 
 
 class PendingAsk(BaseModel):
+    model_config = ConfigDict(arbitrary_types_allowed=True)
+
     user_id: str
     status: Literal["pending", "answered"] = "pending"
     reply: Optional[str] = None
+    event: asyncio.Event = Field(default_factory=asyncio.Event)
```
lilyydu: should we make this a dataclass instead of a BaseModel b/c we're using asyncio here? BaseModel is built for serialization mostly I think

lilyydu (commented):
Added a couple of comments, I like the changes tho :)

packages/api/src/microsoft_teams/api/clients/conversation/client.py
MehakBindra: PR desc was wrong, thanks !

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -130,6 +130,21 @@ def update(self, text: str) -> None:
...
+        )
+        if self._final_activity is not None:
+            self._final_activity.text = None
```
heyitsaamir: Hm, shouldn't we also get rid of all the entities?

packages/apps/src/microsoft_teams/apps/http_stream.py
heyitsaamir, line 147: Hmm, I don't get what the implication of this is. Can you show a gif/video of what the end UX looks like if someone calls clear and sends an AC in the middle?

What kind of behavior is this promoting?

examples/mcp-server/src/mcp_tools.py
MehakBindra: The timeout here isn't an error condition, it's an expected outcome ( timeout happened, event did not occur, return status ). adding a comment for clarity !

examples/mcp-server/src/mcp_tools.py
MehakBindra: I think if we're throwing error, we're good

packages/apps/src/microsoft_teams/apps/http_stream.py
MehakBindra: yes, good catch , this was supposed to mirror TS, dropping all text AND attachments, drifted for some reason

examples/a2a/.env.alice.template
MehakBindra: this is actually dead code, thanks for pointing it out

examples/a2a/src/a2a_client.py
MehakBindra: thanks !

examples/mcp-server/src/models.py
MehakBindra: yes !

packages/apps/src/microsoft_teams/apps/http_stream.py
MehakBindra: good question but don't think so, adding reason in code comment for future clarity

packages/apps/src/microsoft_teams/apps/http_stream.py
MehakBindra: yeah, i think so

packages/apps/src/microsoft_teams/apps/http_stream.py
MehakBindra: stream partial reply as text, then swap the whole thing for a clean card-only result.  clear_text is what prevents the final message from being streamed text + card stacked together, you get just the card.

MehakBindra: uploaded video too

lilyydu (approved):

PR #444, opened 2026-05-28
docs(examples): show anonymous-mode startup warning by default

corinagum:
## Summary

When a PY example app starts without credentials (`CLIENT_ID` / `CLIENT_SECRET` / `TENANT_ID`), the SDK emits a `logger.warning(...)` from `microsoft_teams.apps.http.http_server` telling the developer that the bot will accept unauthenticated requests. Python's root logger has no handler by default, so WARNING records from non-root loggers never reach stdout. A developer running `uv run python src/main.py` therefore sees only uvicorn's INFO lines and is left to discover anonymous mode some other way.

This adds a single `logging.basicConfig(level=logging.INFO)` call near the top of each example's `main.py` so the warning (and any other SDK INFO/WARNING messages) surfaces on stdout alongside uvicorn's output.

### Scope

- `examples/echo` was patched first (commit 1) as the canonical first-run example
- All other examples that did not already configure logging were patched in commit 2: `botbuilder`, `cards`, `meetings`, `message-extensions`, `proactive-messaging`, `quoting`, `reactions`, `stream`, `suggested-actions`, `tab`, `targeted-messages`, `threading`
- Six examples already had their own logging setup and were left untouched: `ai-agentframework`, `dialogs`, `graph`, `mcp-server`, `oauth`

### Before

```
INFO:     Started server process [12345]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:3978 (Press CTRL+C to quit)
```

### After

```
WARNING:microsoft_teams.apps.app:TENANT_ID is not set, assuming multi-tenant app
WARNING:microsoft_teams.apps.http.http_server:No credentials configured (CLIENT_ID / CLIENT_SECRET / TENANT_ID). Bot will accept unauthenticated requests on /api/messages.
INFO:microsoft_teams.apps.app:Teams app initialized successfully
INFO:microsoft_teams.apps.app:Teams app started successfully
INFO:     Started server process [12345]
INFO:     Waiting for application startup.
INFO:     Application startup complete.
INFO:     Uvicorn running on http://0.0.0.0:3978 (Press CTRL+C to quit)
```

## Test plan

- [x] ``uv run poe check`` passes (ruff format + ruff check) across all 13 modified files
- [x] Verified empirically on `examples/echo` with no `.env`: the `No credentials configured ...` WARNING line appears on stdout at startup
- [x] Confirmed uvicorn INFO lines still render unchanged
- [ ] Reviewer to spot-check one of the other patched examples

### DIFF
--- a/examples/botbuilder/src/main.py
+++ b/examples/botbuilder/src/main.py
@@ -5,6 +5,7 @@ Licensed under the MIT License.
 
 import asyncio
 import datetime
+import logging
 import sys
 import traceback
 
@@ -19,6 +20,10 @@ from config import DefaultConfig
 from microsoft_teams.api import MessageActivity
 from microsoft_teams.apps import ActivityContext, App
 from microsoft_teams.botbuilder import BotBuilderPlugin
+
+# Surface SDK INFO/WARNING logs (including the anonymous-mode startup warning
+# emitted when CLIENT_ID / CLIENT_SECRET / TENANT_ID are not configured).
+logging.basicConfig(level=logging.INFO)
 
 config = DefaultConfig()
 adapter = CloudAdapter(ConfigurationBotFrameworkAuthentication(config))
--- a/examples/botbuilder/src/main.py
+++ b/examples/botbuilder/src/main.py
@@ -6,7 +6,6 @@ Licensed under the MIT License.
 import asyncio
 import datetime
 import logging
-import sys
 import traceback
 
 from botbuilder.core import TurnContext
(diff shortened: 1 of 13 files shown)


examples/botbuilder/src/main.py
```diff
@@ -20,6 +21,10 @@
 from microsoft_teams.apps import ActivityContext, App
 from microsoft_teams.botbuilder import BotBuilderPlugin
 
+# Surface SDK INFO/WARNING logs (including the anonymous-mode startup warning
+# emitted when CLIENT_ID / CLIENT_SECRET / TENANT_ID are not configured).
+logging.basicConfig(level=logging.INFO)
```
lilyydu, line 25: did we also want to add
`logger = logging.getLogger(__name__)` to expose the logger per sample?

corinagum: `logging.basicConfig` will surface loggs from any logger including those that come from the Teams SDK itself, while `logging.getLogger` would be for if the sample itself uses the logger. Since it's out of scope for this PR, and the samples often use `print` instead for local logging, I think we can skip in this case.

lilyydu: I'm not a huge fan of using `print` when we have the logger available (esp now that we're exposing it in this PR). I'd prefer just adding the single line to the samples

heyitsaamir (approved):

heyitsaamir (approved):

PR #453, opened 2026-06-06
fix(apps): resolve streaming error handling for 2-minute timeout & content stream not allowed

### LINKED ISSUE #452: HttpStream sends streaming activities faster than the 1 request/second limit
## What happens

When a bot pushes status updates through `HttpStream.update()`, the stream can
send several POSTs back to back, above the 1 request/second limit in the Teams
streaming docs. Teams throttles the stream and ends it: the bot gets a
`403 ContentStreamNotAllowed`, `HttpStream._send` turns it into a
`StreamCancelledError`, and the user sees a red error toast.

It is easiest to hit on the informative `update()` path, for example forwarding
workflow or agent progress events while a longer answer is generated.

- Package: `microsoft-teams-apps` 2.0.11 (same code on `main` / 2.0.13-dev)
- File: `packages/apps/src/microsoft_teams/apps/http_stream.py`

The docs are explicit: "The throttling limit is 1 request per second", make
subsequent calls only after a successful response, and "You can set only one
informative message. Your bot reuses this message for each update."

## How to reproduce

Drive `HttpStream` with a mock client that records the time of each POST and
simulates a ~20 ms round trip:

```python
# A) eight informative updates in a tight loop (bursty progress)
for i in range(8):
    stream.update(f"progress {i}")
await asyncio.sleep(1.0)

# B) five informative updates, 200 ms apart (steady progress)
for i in range(5):
    stream.update(f"step {i}")
    await asyncio.sleep(0.2)
```

```
A) eight rapid update():  8 POSTs over 153 ms  -> ~52 req/s   (limit is 1)
B) five update() @ 200ms: 5 POSTs, min gap 201 ms             (should be >= 1000 ms)
```

Both go well past 1 req/s, enough to get the stream throttled against the real
service.

## Why it happens

Two separate causes in `http_stream.py`.

First, `_flush()` drains the queue and sends each collected informative update in
a loop with nothing between the calls, then the text chunk immediately after:

```python
for typing_update in informative_updates:
    await self._send_activity(typing_update)   # no delay between calls
if self._text:
    await self._send_activity(TypingActivityInput(text=self._text))  # also immediate
```

So eight queued `update()` calls become eight POSTs in one flush. The text path
is already coalesced into a single chunk, so only the informative path bursts.

Second, the only pacing is `call_later(0.5, ...)` (already 2 req/s), armed only
when a backlog survives a flush:

```python
if self._queue and not self._timeout:
    self._timeout = asyncio.get_running_loop().call_later(0.5, ...)
```

Once a flush empties the queue, `_timeout` stays `None`, so the next `emit()`
schedules an immediate flush and a steady stream of `update()` calls gets no
pacing at all. That is case B.

## Possible fix

The throttle is better handled inside `HttpStream` than in every caller, and #453
does that with a per-stream limiter. The SDK creates one `HttpStream` per
conversation turn, so a stream-scoped limiter matches the per-conversation scope
Teams throttles on. Two parts:

1. Pace all sends to at most 1 req/s, including across flushes. The first send
   goes out right away so the progress bar appears quickly; the rest wait for the
   interval, which is configurable toward the docs' 1.5 to 2s advice.
2. By default coalesce a burst of informative updates in one flush to the latest,
   matching the docs' one-reused-message behavior. A flag paces out every update
   instead, for callers who want each one delivered.

## Environment

- microsoft-teams-apps 2.0.11, Python 3.12
- msteams channel, 1:1 chat, streaming

Athosone:
PR has evolved due to a number of raised issues wrt streaming error handling. This PR solely focuses on handling the 403 for the 2-minute timeout, and surfacing the "Content stream is not allowed" error.

This table highlights all the different error codes, we are concerned with the 403s:
https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux?tabs=csharp#error-codes

1) Content stream is not allowed -> Error is surfaced, not as a cancelled error
2) Content stream is not allowed on an already completed streamed message -> Not possible with the SDK
3) Content stream finished due to exceeded streaming time. -> Handled in this PR. A new regular message is sent instead.
4) Content stream was canceled by user -> Already handled today
5) Request streamed content should contain the previously streamed content -> Not possible with the SDK

**TO BE DISCUSSED:**
1) Bursty informative updates - continued discussion in #452 to reach an ideal solution
2) Message size too large -> To be discussed in #488

---
OLD DESCRIPTION:

Fixes #452

## Problem

`HttpStream` can send streaming activities faster than the Teams 1 request/second
per-stream limit. Teams then throttles the stream and returns
`403 ContentStreamNotAllowed`, which `_send` converts to `StreamCancelledError`
and the user sees as a red error toast.

Two causes in `http_stream.py`:

1. `_flush()` sent every queued informative update back to back, then the text
   chunk. A burst of `update()` calls became a burst of POSTs in one flush.
2. The only pacing was a `call_later(0.5, ...)` reschedule, armed only when a
   backlog survived a flush. Once the queue drained, `_timeout` was `None`, so the
   next `emit()` fired an immediate flush with no pacing.

## Fix

A per-stream leaky-slot limiter (`make_limiter` in `utils/limiter.py`), gating
every chunk send in `_send_activity`. Its `next_slot` is instance state shared
across flushes, so it paces both the in-flush burst and the next `emit()` after
the queue drains. The first send is not delayed, so the progress bar still
appears promptly.

- `min_send_interval` on `HttpStream.__init__` (default 1.0s) lets callers buffer
  toward the docs' 1.5 to 2s advice.
- `coalesce_informative_updates` defaults to `True`: a burst of informative
  updates in one flush collapses to the latest, matching the docs' "one
  informative message, reused for each update". Set it `False` to pace out every
  update at 1 req/s instead (see Notes).

## Tests

Limiter spacing and input validation; rapid `update()` burst coalesces to the
latest by default; with `coalesce_informative_updates=False` every update is sent
in order, none dropped, paced at the interval; emits are paced across flushes (the
post-drain case); the interval is configurable; the flag never drops text.
Existing `test_http_stream.py` behaviour is unchanged. Full `packages/apps` suite
passes, ruff and pyright clean.

## Notes

- The limiter is acquired in `_send`, so every HTTP attempt is paced: retries and
  `close()`'s final send included. `acquire()` only waits when a send would
  actually be too soon, so close pays no latency unless it lands within the
  interval of the last chunk. Pass `min_send_interval=0` to disable pacing.
- With `coalesce_informative_updates=False`, `_flush` holds its lock across the
  paced sleeps, so a very long informative burst could keep the lock long enough
  that `close()` times out. The default (coalesce on) avoids this.
- The same gap likely exists in the sibling TS and .NET SDKs; not addressed here.


Athosone:
@microsoft-github-policy-service agree company="Michelin"

packages/apps/src/microsoft_teams/apps/utils/limiter.py
```diff
@@ -0,0 +1,30 @@
...
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+import asyncio
+from time import monotonic
+from typing import Awaitable, Callable
+
+
+def make_limiter(rate: int, period: float = 1.0) -> Callable[[], Awaitable[None]]:
```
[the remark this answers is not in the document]
Athosone: Already handled: make_limiter raises ValueError for rate < 1 and period < 0, so rate == 0 (ZeroDivisionError) and negative periods fail fast at construction. period == 0 is allowed on purpose and means "no pacing" (interval 0), which is how callers disable the limiter.

packages/apps/src/microsoft_teams/apps/utils/limiter.py
```diff
@@ -0,0 +1,30 @@
...
+    """Leaky-slot limiter: at most `rate` acquisitions per `period` seconds.
+
+    The slot is reserved (read then write of `next_slot`) with no await in
+    between, so reservations are race-free under single-threaded asyncio. The
+    first call never waits.
+    """
+    interval = period / rate
```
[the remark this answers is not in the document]
Athosone: Same as above: rate < 1 and period < 0 already raise ValueError in make_limiter, so the interval can never be derived from a zero rate or a negative period. period == 0 is intentionally allowed to mean no pacing.

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -278,6 +294,7 @@ async def _send_activity(self, to_send: TypingActivityInput):
         Args:
             activity: The activity to send.
         """
+        await self._acquire()
```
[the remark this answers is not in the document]
Athosone: Good catch, fixed in bd1e0a3. The acquire() call moved from _send_activity into _send, so every HTTP attempt is paced: retries and the final close() send included. acquire() only waits when a send would actually be too soon (next_slot is in the past after any idle gap), so close() pays no latency unless it lands within the interval of the last chunk. Added test_close_final_send_is_paced to cover it.

packages/apps/tests/test_http_stream.py
```diff
@@ -453,3 +454,93 @@ async def test_close_waits_for_flush_to_complete(self, mock_api_client, conversa
...
+
+        mock_api_client.conversations.activities().create = mock_send
+
+        stream = HttpStream(mock_api_client, conversation_reference, min_send_interval=interval)
+        for i in range(8):
+            stream.update(f"progress {i}")
+
+        task = stream._pending
+        assert task is not None
+        await task
```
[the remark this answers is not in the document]
Athosone: Leaving these as-is. Awaiting the flush task is the synchronization point these tests need, and there is no public drain/flush hook today. Adding one purely for tests would put test-only surface into the production class. If a public drain helper lands later for other reasons, happy to migrate these asserts onto it.

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -32,27 +32,44 @@ class HttpStream(StreamerProtocol):
...
+    out every update instead.
     """
 
-    def __init__(self, client: ApiClient, ref: ConversationReference):
+    def __init__(
+        self,
+        client: ApiClient,
+        ref: ConversationReference,
+        min_send_interval: float = 1.0,
+        coalesce_informative_updates: bool = True,
```
lilyydu: `activity_sender.py` constructs `HttpStream` without passing `min_send_interval` or `coalesce_informative_updates`. We should also plumb these through activity_sender / app config

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -32,27 +32,44 @@ class HttpStream(StreamerProtocol):
...
+    ):
         """
         Initialize a new HttpStream instance.
 
         Args:
             client (ApiClient): The API client used to send activities to Microsoft Teams.
             ref (ConversationReference): Reference to the Teams conversation.
+            min_send_interval (float): Minimum seconds between sends, including retries and the final
+                close() send (Teams limits streaming to 1 req/s). Set 0 to disable pacing.
+            coalesce_informative_updates (bool): When True (default), a burst of informative updates in
```
lilyydu: hmm setting this to true means itll silently drop intermediate informative bubbles within a flush

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -32,27 +32,44 @@ class HttpStream(StreamerProtocol):
...
+                one flush collapses to the latest one. Set False to pace out every update at 1 req/s
+                instead; a long burst then holds the flush lock and can delay or drop close()'s final
+                message (see _total_wait_timeout).
         """
         super().__init__()
         self._client = client
         self._ref = ref
         self._events = EventEmitter[StreamerEvent]()
+        self._acquire = make_limiter(rate=1, period=min_send_interval)
```
lilyydu: nit: should we just do `make_limiter(interval: float)`?

packages/apps/src/microsoft_teams/apps/utils/limiter.py
```diff
@@ -0,0 +1,35 @@
...
+Licensed under the MIT License.
+"""
+
+import asyncio
+from time import monotonic
+from typing import Awaitable, Callable
+
+
+def make_limiter(rate: int, period: float = 1.0) -> Callable[[], Awaitable[None]]:
+    """Leaky-slot limiter: at most `rate` acquisitions per `period` seconds.
```
lilyydu: nit: can we rephrase to fixed-interval gate / token bucket of size 1

Athosone:
Thanks a lot for taking the time to review @lilyydu I will come back to you soon.

Btw I noticed two other issues while playing with it:
1. 2 minutes timeout hard-coded for the streaming response. This is a problem for us, we are running a chat bot that helps our support user to resolve incident, doing RCA etc. sometimes it can takes some times to gather all the required info, would you mind if I proposed a new issue/pr to address this? The fix would be to allow the user to configure the timeout via the app config?
2. When we send a payload to the stream if it is too big we receive a 413 entity too large, The fix I would like to propose is that the httpstream chunk the payload and send it in batch of "MAX_SIZE". I can also make a PR/issue if you are ok with it?

However I wonder if the httpstream should carry that responsability alone, are these constraints only tight to it?

packages/apps/src/microsoft_teams/apps/http_stream.py
Athosone: Done in fa6ae49. Added `stream_min_send_interval` and `stream_coalesce_informative_updates` to AppOptions, App passes them into ActivitySender, and create_stream() forwards them to HttpStream.

packages/apps/src/microsoft_teams/apps/http_stream.py
Athosone: Yeah, agreed. Flipped the default to False in fa6ae49: every informative update is kept and paced at min_send_interval, and coalescing is opt-in. This also matches what the stream did before this PR, where every informative update was sent. One trade-off to be aware of: a long burst now holds the flush lock for roughly burst_size * interval seconds, so it can delay the final close() send (bounded by _total_wait_timeout). The docstring covers this.

packages/apps/src/microsoft_teams/apps/http_stream.py
Athosone: Done

packages/apps/src/microsoft_teams/apps/utils/limiter.py
Athosone: Done in fa6ae49. Now reads: "Fixed-interval gate (a token bucket of size 1): consecutive acquisitions are spaced at least `interval` seconds apart."

packages/apps/src/microsoft_teams/apps/activity_sender.py
```diff
@@ -27,14 +27,25 @@ class ActivitySender:
     Separate from transport concerns (HTTP, WebSocket, etc.)
     """
 
-    def __init__(self, client: Client):
+    def __init__(
+        self,
+        client: Client,
+        stream_min_send_interval: float = 1.0,
+        stream_coalesce_informative_updates: bool = False,
```
heyitsaamir: Let's stream_coalesce_informative_updates should be true always. let's not expose this.

packages/apps/src/microsoft_teams/apps/activity_sender.py
```diff
@@ -27,14 +27,25 @@ class ActivitySender:
     Separate from transport concerns (HTTP, WebSocket, etc.)
     """
 
-    def __init__(self, client: Client):
+    def __init__(
+        self,
+        client: Client,
+        stream_min_send_interval: float = 1.0,
```
heyitsaamir: also don't think we need this.

heyitsaamir:
Thanks for the push here!

Some thoughts:
1. Let’s not expose configurable flags for this. Informative updates should be coalesced by default because they’re status replacements, not cumulative content. If we skip intermediate informative updates, let’s log only the count at debug level.

2. I think HttpStream should have two modes:
   - informative mode: before text streaming starts, send the latest pending informative update
   - text mode: once text streaming starts, do not send informative updates anymore

Can we still use the same retry path and the same send pacing/limiter for informative updates, text chunks, retries, and final close.

heyitsaamir:
> Thanks a lot for taking the time to review @lilyydu I will come back to you soon.
> 
> Btw I noticed two other issues while playing with it:
> 
> 1. 2 minutes timeout hard-coded for the streaming response. This is a problem for us, we are running a chat bot that helps our support user to resolve incident, doing RCA etc. sometimes it can takes some times to gather all the required info, would you mind if I proposed a new issue/pr to address this? The fix would be to allow the user to configure the timeout via the app config?
> 2. When we send a payload to the stream if it is too big we receive a 413 entity too large, The fix I would like to propose is that the httpstream chunk the payload and send it in batch of "MAX_SIZE". I can also make a PR/issue if you are ok with it?
> 
> However I wonder if the httpstream should carry that responsability alone, are these constraints only tight to it?

Yeah these are outlined in the [official streaming docs](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux?tabs=csharp). Admittedly, the streaming system here _could_ help by just closing the stream and just doing regular updates.
Were you thinking something along the same lines?

I reached out to the folks that implemented streaming to ask what their preferred recommendation would be for these cases.

Athosone:
Hi !
Thanks for your feedback,

If this is a platform limit that you cannot increase then yes we could fallback updating the message (this is what I am doing on my app at the moment but the experience is not as nice as with a stream :p).

For your comment on this PR I totally agree and will update the code, removing the flag in app config and coalesce the informative update and applying the limiter to different code path that emit bytes to the platform.

It matches my experience in the sense of, the user doesnt really care for the next narrative update of an agent or the result of a tool call if you already have the answer ready to be sent.

lilyydu:
Hey @Athosone,

We're still waiting on guidance from internal folks, but I'm gonna test out your implementation in the meantime. Thanks for responding to our requests!

lilyydu:
Hey @Athosone,

Can you share what error message you get back for the 403 you are receiving from the 1 request /sec limitation? There's a couple that fall into that bucket here https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux?tabs=csharp#error-codes and want to make sure we cover our bases

lilyydu:
> Hey @Athosone,
> 
> Can you share what error message you get back for the 403 you are receiving from the 1 request /sec limitation? There's a couple that fall into that bucket here https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux?tabs=csharp#error-codes and want to make sure we cover our bases

@Athosone soft bump so we can unblock you!

Athosone:
SOrry about that, totally missed the notification,

I got different error messages:
```
The bot failed to complete the streaming process within the strict time limit of two minutes
```

And 

```
The bot sent a message that exceeds the current [message size](https://learn.microsoft.com/en-us/microsoftteams/platform/bots/how-to/format-your-bot-messages) restriction
```

For the first one I morph the stream to a message

For the second I implemented a chunker so that my adapter abstract the complexity to my main code.

lilyydu:
Hey @Athosone ,

First, thanks for all the investigation and work you've done here! Streaming can definitely be improved so we highly appreciate your feedback and suggestions. A number of issues have been raised so we think it's best to isolate each into its own respective PR

In regard to the bursty informative updates, I repro'd the code snippets you attached, and it does result in 429s but these are handled via our retry logic. However, we do agree that we should have some mechanism to control the bursty behaviour. I tried out the fix you're proposing here with 100 informative updates but it coalesced into one single msg - not sure if that's the ideal experience too. Let's continue bouncing ideas in #452 for this

The 403s you were receiving are due to the 2 errors you attached above (2 minute limitation and chunk size). We currently return all 403s as "cancelled" which does not properly showcase the error table I linked above. I'm going to update this PR to surface the 2 minute error instead and then open up additional PRs following after to surface the other errors.

Apologies for the delay with resolving this but glad we are surfacing and inching closer to a better streaming exp :)

lilyydu:
Updated the description:

PR has evolved due to a number of raised issues wrt streaming error handling. This PR solely focuses on handling the 403 for the 2-minute timeout, and surfacing the "Content stream is not allowed" error.

This table highlights all the different error codes, we are concerned with the 403s:
https://learn.microsoft.com/en-us/microsoftteams/platform/bots/streaming-ux?tabs=csharp#error-codes

1. Content stream is not allowed -> Error is surfaced, not as a cancelled error
2. Content stream is not allowed on an already completed streamed message -> Not possible with the SDK
3. Content stream finished due to exceeded streaming time. -> Handled in this PR. A new regular message is sent instead.
4. Content stream was canceled by user -> Already handled today
5. Request streamed content should contain the previously streamed content -> Not possible with the SDK


**TO BE DISCUSSED:**

1. Bursty informative updates - continued discussion in https://github.com/microsoft/teams.py/issues/452 to reach an ideal solution
2. Message size too large -> To be discussed in #488

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -315,6 +348,13 @@ async def _send(self, to_send: Union[TypingActivityInput, MessageActivityInput])
             return SentActivity.merge(to_send, res)
         except HTTPStatusError as e:
             if e.response.status_code == 403:
+                error = e.response.json().get("error", {})
+                message = error.get("message", "")
+                if message != "Content stream was cancelled by user.":
+                    if message == "Content stream finished due to exceeded streaming time.":
+                        self._timed_out = True
+                    logger.warning("Teams encountered an error while streaming. Sending as a regular message.")
+                    raise StreamCancelledError(message) from e
```
heyitsaamir: This will raise the StreamCancelled error to the handler. We should probably raise a different error here.

Can we just do a switch statement here for various types of canceled errors? Also point to the learn docs that describe these.

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -315,6 +348,13 @@ async def _send(self, to_send: Union[TypingActivityInput, MessageActivityInput])
             return SentActivity.merge(to_send, res)
         except HTTPStatusError as e:
             if e.response.status_code == 403:
+                error = e.response.json().get("error", {})
+                message = error.get("message", "")
+                if message != "Content stream was cancelled by user.":
```
heyitsaamir: let's also make sure that the backend has tests for these particular strings

heyitsaamir: (and maybe a more resilient error could be to just check for phrases like "canceled", "exceeded streaming time" etc.)

Athosone:
> Hey @Athosone ,
> 
> 
> 
> First, thanks for all the investigation and work you've done here! Streaming can definitely be improved so we highly appreciate your feedback and suggestions. A number of issues have been raised so we think it's best to isolate each into its own respective PR
> 
> 
> 
> In regard to the bursty informative updates, I repro'd the code snippets you attached, and it does result in 429s but these are handled via our retry logic. However, we do agree that we should have some mechanism to control the bursty behaviour. I tried out the fix you're proposing here with 100 informative updates but it coalesced into one single msg - not sure if that's the ideal experience too. Let's continue bouncing ideas in #452 for this
> 
> 
> 
> The 403s you were receiving are due to the 2 errors you attached above (2 minute limitation and chunk size). We currently return all 403s as "cancelled" which does not properly showcase the error table I linked above. I'm going to update this PR to surface the 2 minute error instead and then open up additional PRs following after to surface the other errors.
> 
> 
> 
> Apologies for the delay with resolving this but glad we are surfacing and inching closer to a better streaming exp :)

No problem and I am glad too that this topic move forward.
I am on vacation hence the delay but I will keep following it :)

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -194,11 +210,20 @@ async def close(self) -> Optional[SentActivity]:
...
+            activity.with_text(self._text).with_channel_data(self._channel_data)
+            activity.id = self._id
+            activity.add_stream_final()
+            try:
+                res = await self._send_with_retry(activity)
+            except StreamTimedOutError:
+                # The final streamed send tripped the 2-minute limit. Update the original
+                # message in place with the buffered content: reuse the id and drop the
+                # streamInfo entity + stream channel data so this routes to update, not create.
+                res = await self._send_final()
```
heyitsaamir, line 226: @lilyydu i did a mini refactor here b/c there was a lot of reused code.

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -314,8 +371,32 @@ async def _send(self, to_send: Union[TypingActivityInput, MessageActivityInput])
...
             if e.response.status_code == 403:
-                self._canceled = True
-                logger.warning("Teams channel stopped the stream.")
-                raise StreamCancelledError("Teams channel stopped the stream.") from e
+                message = ""
+                try:
+                    message = e.response.json().get("error", {}).get("message", "")
+                except ValueError:
+                    # 403 with an empty or non-JSON body: treat as an unknown stream error.
+                    pass
```
heyitsaamir, line 382: Should this be re-raised?

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -210,6 +235,32 @@ async def close(self) -> Optional[SentActivity]:
...
+    async def _send_with_retry(
+        self,
+        activity: Union[TypingActivityInput, MessageActivityInput],
+        options: Optional[RetryOptions] = None,
+    ) -> SentActivity:
+        """Send an activity through retry, treating terminal stream errors as non-retryable."""
+        return await retry(
+            lambda: self._send(activity),
+            options=options or RetryOptions(),
+            non_retryable=(TerminalStreamError,),
```
heyitsaamir, line 247: How come `StreamNotAllowedError` isn't non-retryable?

lilyydu: reread your question - I don't think we should try again cause StreamNotAllowedError means the streaming API isn't allowed for the user or bot

packages/apps/src/microsoft_teams/apps/http_stream.py
lilyydu: it does get re-raised as TerminalStreamError

packages/apps/src/microsoft_teams/apps/http_stream.py
heyitsaamir: Right, so shouldn't it be in this array?

heyitsaamir (approved):

lilyydu (approved):

PR #455, opened 2026-06-09
fix: parse attachments in MessageUpdateActivity (#414)

### LINKED ISSUE #414: [Bug]: attachments are not parsed on message update
### Bug Description

Attachments look like this on a new message activity:

```
attachments=[Attachment(content_type='text/html', content_url=None, content='<p>hey</p>', name=None, thumbnail_url=None)]
```

But on message update it looks like this:
```
attachments=[{'contentType': 'text/html', 'content': '<p>hey2</p>'}]
```

### Steps to Reproduce

1. Send message
2. Update message

### Expected Behavior

Activities are parsed in the same way for all activity types.

### Actual Behavior

Activies are not parsed in the same way for all activity types.

### SDK Version

2.0.0a78

### Python Version

3.14.2

### Additional Context

_No response_

dcaayushd:
Fixes #414

attachments field was missing from MessageUpdateActivity so
edited messages with attachments silently dropped them. added
the same Optional[List[Attachment]] field that MessageActivity has.

added a regression test.

### DIFF
--- a/packages/api/src/microsoft_teams/api/activities/message/message_update.py
+++ b/packages/api/src/microsoft_teams/api/activities/message/message_update.py
@@ -4,9 +4,9 @@ Licensed under the MIT License.
 """
 
 from datetime import datetime
-from typing import Any, Literal, Optional
+from typing import Any, List, Literal, Optional
 
-from ...models import ActivityBase, ChannelData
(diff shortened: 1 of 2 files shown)


lilyydu (approved):
Thank you for fixing!

PR #475, opened 2026-06-16
test: add integration tests for Teams Python SDK

corinagum:
Adds 12 integration tests that make real API calls against the Teams Bot Framework service.

**Tests:**
- Activities: create, update, reply, delete (4)
- Members: get all, get by ID, get paged (3)
- Conversations: create 1:1, create group (2)
- Teams: get details, get channels (2)
- Reactions: add and delete (1)

**Also fixes:**
- `TeamDetails.type` was incorrectly marked as required — service doesn't always return it. Now `Optional`.

**Setup:**
```bash
uv sync --all-packages --group dev --group integration
export $(cat tests/integration/.env.botid-prod | xargs)
pytest tests/integration -v
```

See [cross-SDK runbook](https://github.com/microsoft/teams-sdk/blob/main/INTEGRATION-TESTS.md) for provisioning and troubleshooting.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -123,8 +123,9 @@ async def fixture():
         members = await api.conversations.members(config.conversation_id).get_all()
         bot_prefix = f"28:{config.client_id}"
         _cached_members = [m for m in members if m.id and not m.id.startswith(bot_prefix)]
-        if _cached_members:
-            _cached_member_mri_1 = _cached_members[0].id
+        if not _cached_members:
+            raise RuntimeError("No non-bot members found in conversation — tests require at least one human member")
+        _cached_member_mri_1 = _cached_members[0].id
         if len(_cached_members) > 1:
             _cached_member_mri_2 = _cached_members[1].id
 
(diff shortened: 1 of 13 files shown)


tests/integration/conftest.py
```diff
@@ -0,0 +1,138 @@
...
+        service_url=env("TEST_SERVICE_URL"),
+        conversation_id=env("TEST_CONVERSATION_ID"),
+        user_id=env("TEST_USER_ID"),
+        team_id=env("TEST_TEAM_ID"),
+        channel_id=env("TEST_CHANNEL_ID"),
+        meeting_id=env("TEST_MEETING_ID"),
+        user_id_2=os.environ.get("TEST_USER_ID_2"),
+        agentic_app_id=os.environ.get("TEST_AGENTIC_APP_ID"),
+        agentic_user_id=os.environ.get("TEST_AGENTIC_USER_ID"),
+        scope=os.environ.get("AZURE_SCOPE", "https://api.botframework.com/.default"),
```
heyitsaamir, line 94: One idea is to have a script that helps us create this (using teamscli + az maybe)

heyitsaamir (approved):

tests/integration/conftest.py
corinagum: yep, definitely on the list!

PR #464, opened 2026-06-11
Agent 365 support

heyitsaamir:
This is the Python feature branch for Agent 365 and ACF support in 2.1. It brings the stacked identity, lifecycle, proactive messaging, and observability work together as one integration point.

## Included commits

- [`24b0ffa`](https://github.com/microsoft/teams.py/commit/24b0ffa9ccf9b61617328fbba20215ac20fec898) Start Agent 365 support branch
- [`ecd2654`](https://github.com/microsoft/teams.py/commit/ecd265454fb24e80826b13b552658561c39e39e7) Add Agent 365 token support (#457)
- [`02c2091`](https://github.com/microsoft/teams.py/commit/02c20914d395a1a027ccaf2531cdffccd1830e0e) Support Entra inbound Agent ID tokens (#462)
- [`6198553`](https://github.com/microsoft/teams.py/commit/61985539d69b3f1ae1c7bad40cf092144625e1fc) Add agentic auth provider infrastructure (#485)
- [`2f8b767`](https://github.com/microsoft/teams.py/commit/2f8b767934c0e9654da290c222a19efe8aaf5094) Add AgenticIdentity API surface (#478)
- [`5ce920a`](https://github.com/microsoft/teams.py/commit/5ce920a2b671f6032690030e1578537a2a4ec70e) Remove ActivitySender send abstraction (#479)
- [`b65a997`](https://github.com/microsoft/teams.py/commit/b65a9976d5a6f15a5f6c4130ce9f0665824e6a40) Add app-level agentic send helpers (#480)
- [`3b441b8`](https://github.com/microsoft/teams.py/commit/3b441b8a6ece38f014a873c787de4205c284ced2) Default agentic app blueprint ID to app client ID
- [`11a9bb8`](https://github.com/microsoft/teams.py/commit/11a9bb861846bdff754ba0a5a611755507b6f960) Add Agent 365 agentLifecycle event support (#492)
- [`b5b1d44`](https://github.com/microsoft/teams.py/commit/b5b1d44be85890a6df8e33a41fde24d95eae655d) Fix Agent365 base branch deprecation errors
- [`36e7c27`](https://github.com/microsoft/teams.py/commit/36e7c27bba891eee4532664ed98cdd3ad1d47b06) Simplify API client scoping (#505)
- [`b620848`](https://github.com/microsoft/teams.py/commit/b6208480d7794f0ac13d86381d78cfce36c5827d) Add OpenTelemetry API middleware (#514)
[...]

PR #478, opened 2026-06-17
Add AgenticIdentity API surface

heyitsaamir:
This PR adds `AgenticIdentity` as a new auth context across the Teams API surface.

Why:
Agent ID calls need a user-shaped Agent ID token instead of the normal bot token. The API methods now accept `agentic_identity` where needed, and the shared auth interceptor uses local request metadata to pick the right auth path.

Interesting bits:
- Adds optional `agentic_identity` to conversation create, activity operations, conversation members, reactions, teams, and meetings.
- Adds optional `service_url` overrides across the service-url based APIs touched here.
- Agentic identity is passed through `httpx` request extensions, not headers. So it stays inside the client pipeline and does not leak to Teams endpoints.
- The auth interceptor reads that extension and asks the auth provider for the right token.

Reviewer tips:
Start with the API method signatures, then the changed call sites in `ConversationActivityClient`, `ConversationMemberClient`, `ReactionClient`, `TeamClient`, and `MeetingClient`. The auth plumbing itself lives one PR down in #485.

Testing:
- Focused API unit tests for conversation, reaction, bot, user, team, and meeting clients.
- Full test suite passed locally.

Live smoke tested with Agent ID token:
- conversation activity create/update/reply/get_members
- conversation members get_all/get/get_paged
- reactions add/delete
- teams get_by_id/get_conversations
- conversations.create
- meetings get_by_id/get_participant

Known limitation:
- targeted activity create/update reached the service but returned 500. Leaving that as not proven until platform behavior is confirmed.

### DIFF
--- a/packages/api/src/microsoft_teams/api/clients/base_client.py
+++ b/packages/api/src/microsoft_teams/api/clients/base_client.py
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 Licensed under the MIT License.
 """
 
-from typing import Optional, Union
+from typing import Optional, Union, cast
 
 from microsoft_teams.common import Client, ClientOptions
 
@@ -21,7 +21,7 @@ class BaseClient:
         """Initialize the BaseClient.
 
         Args:
-            options: Optional Client or ClientOptions instance. If not provided, a default Client will be created.
+            options: Optional Client or ClientOptions instance. If not provided, a default Client is created.
             api_client_settings: Optional API client settings.
         """
         if options is None:
@@ -42,3 +42,10 @@ class BaseClient:
     def http(self, value: Client) -> None:
         """Set the HTTP client instance."""
         self._http = value
+
+    def _get_service_url(self, service_url: str | None = None) -> str:
+        current_service_url = cast(str | None, getattr(self, "service_url", None))
+        resolved_service_url = service_url or current_service_url
+        if resolved_service_url is None:
+            raise ValueError("service_url is required")
+        return resolved_service_url.rstrip("/")
--- a/packages/api/src/microsoft_teams/api/clients/conversation/activity.py
+++ b/packages/api/src/microsoft_teams/api/clients/conversation/activity.py
@@ -9,7 +9,8 @@ from microsoft_teams.common.experimental import experimental
 from microsoft_teams.common.http import Client
 
 from ...activities import ActivityParams, SentActivity
-from ...models import TeamsChannelAccount
+from ...models import AgenticIdentity, TeamsChannelAccount
+from .._auth_provider_interceptor import AGENTIC_IDENTITY_EXTENSION
 from ..api_client_settings import ApiClientSettings
 from ..base_client import BaseClient
 
@@ -38,7 +39,14 @@ class ConversationActivityClient(BaseClient):
         super().__init__(http_client, api_client_settings)
         self.service_url = service_url.rstrip("/")
 
-    async def create(self, conversation_id: str, activity: ActivityParams) -> SentActivity:
+    async def create(
+        self,
+        conversation_id: str,
+        activity: ActivityParams,
+        *,
+        service_url: str | None = None,
+        agentic_identity: AgenticIdentity | None = None,
+    ) -> SentActivity:
         """
         Create a new activity in a conversation.
 
@@ -51,8 +59,9 @@ class ConversationActivityClient(BaseClient):
         """
 
         response = await self.http.post(
-            f"{self.service_url}/v3/conversations/{conversation_id}/activities",
+            f"{self._get_service_url(service_url)}/v3/conversations/{conversation_id}/activities",
             json=activity.model_dump(by_alias=True, exclude_none=True),
+            extensions={AGENTIC_IDENTITY_EXTENSION: agentic_identity},
         )
 
         # Note: Typing activities (non-streaming) always produce empty responses.
@@ -61,7 +70,15 @@ class ConversationActivityClient(BaseClient):
         id = response.json().get("id", _PLACEHOLDER_ACTIVITY_ID)
         return SentActivity(id=id, activity_params=activity)
 
-    async def update(self, conversation_id: str, activity_id: str, activity: ActivityParams) -> SentActivity:
+    async def update(
+        self,
+        conversation_id: str,
+        activity_id: str,
+        activity: ActivityParams,
+        *,
+        service_url: str | None = None,
+        agentic_identity: AgenticIdentity | None = None,
+    ) -> SentActivity:
         """
         Update an existing activity in a conversation.
 
@@ -74,13 +91,22 @@ class ConversationActivityClient(BaseClient):
             The updated activity
         """
         response = await self.http.put(
-            f"{self.service_url}/v3/conversations/{conversation_id}/activities/{activity_id}",
+            f"{self._get_service_url(service_url)}/v3/conversations/{conversation_id}/activities/{activity_id}",
             json=activity.model_dump(by_alias=True, exclude_none=True),
+            extensions={AGENTIC_IDENTITY_EXTENSION: agentic_identity},
         )
         id = response.json()["id"]
         return SentActivity(id=id, activity_params=activity)
 
-    async def reply(self, conversation_id: str, activity_id: str, activity: ActivityParams) -> SentActivity:
+    async def reply(
+        self,
+        conversation_id: str,
+        activity_id: str,
+        activity: ActivityParams,
+        *,
+        service_url: str | None = None,
+        agentic_identity: AgenticIdentity | None = None,
+    ) -> SentActivity:
         """
         Reply to an activity in a conversation.
 
@@ -95,13 +121,21 @@ class ConversationActivityClient(BaseClient):
         activity_json = activity.model_dump(by_alias=True, exclude_none=True)
         activity_json["replyToId"] = activity_id
         response = await self.http.post(
-            f"{self.service_url}/v3/conversations/{conversation_id}/activities/{activity_id}",
+            f"{self._get_service_url(service_url)}/v3/conversations/{conversation_id}/activities/{activity_id}",
             json=activity_json,
+            extensions={AGENTIC_IDENTITY_EXTENSION: agentic_identity},
         )
         id = response.json()["id"]
         return SentActivity(id=id, activity_params=activity)
 
-    async def delete(self, conversation_id: str, activity_id: str) -> None:
+    async def delete(
+        self,
+        conversation_id: str,
+        activity_id: str,
+        *,
+        service_url: str | None = None,
+        agentic_identity: AgenticIdentity | None = None,
+    ) -> None:
         """
         Delete an activity from a conversation.
 
@@ -109,9 +143,19 @@ class ConversationActivityClient(BaseClient):
             conversation_id: The ID of the conversation
             activity_id: The ID of the activity to delete
         """
-        await self.http.delete(f"{self.service_url}/v3/conversations/{conversation_id}/activities/{activity_id}")
+        await self.http.delete(
+            f"{self._get_service_url(service_url)}/v3/conversations/{conversation_id}/activities/{activity_id}",
+            extensions={AGENTIC_IDENTITY_EXTENSION: agentic_identity},
+        )
 
-    async def get_members(self, conversation_id: str, activity_id: str) -> List[TeamsChannelAccount]:
+    async def get_members(
+        self,
+        conversation_id: str,
+        activity_id: str,
+        *,
+        service_url: str | None = None,
+        agentic_identity: AgenticIdentity | None = None,
+    ) -> List[TeamsChannelAccount]:
         """
         Get the members associated with an activity.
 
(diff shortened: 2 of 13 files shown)


heyitsaamir:

- #464
    - #478 ◀
        - #479
            - #480

packages/api/src/microsoft_teams/api/auth/cloud_environment.py
```diff
@@ -68,6 +73,7 @@ class CloudEnvironment:
     login_endpoint="https://login.partner.microsoftonline.cn",
     login_tenant="microsoftservices.partner.onmschina.cn",
     bot_scope="https://api.botframework.azure.cn/.default",
+    agentic_bot_scope="https://botapi.skype.com/.default",
```
heyitsaamir, line 79: // TODO: need to figur eout the values for other clouds here.

packages/api/src/microsoft_teams/api/models/message/message.py
```diff
@@ -56,6 +56,11 @@ class Message(CustomBaseModel):
     from_: Optional[MessageFrom] = None
     "Sender of the message."
 
+    @property
```
heyitsaamir: let's remove this.

packages/apps/src/microsoft_teams/apps/token_manager.py
```diff
@@ -238,6 +263,30 @@ async def _get_token_with_token_provider(
...
+        scope: str,
+        tenant_id: str,
+        agentic_identity: AgenticIdentity | None = None,
+    ) -> str | Awaitable[str]:
+        token_provider = cast(Any, credentials.token)
+        try:
+            parameters = list(signature(token_provider).parameters.values())
+        except (TypeError, ValueError):
+            return cast(str | Awaitable[str], token_provider(scope, tenant_id, agentic_identity=agentic_identity))
+
```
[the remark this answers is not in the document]
heyitsaamir: agree

packages/api/src/microsoft_teams/api/clients/base_client.py
```diff
@@ -42,3 +58,16 @@ def http(self) -> Client:
...
+
+    def _get_agentic_token(self, agentic_identity: AgenticIdentity | None = None):
+        identity = self._resolve_agentic_identity(agentic_identity)
+        if identity is None:
+            return None
+        if self._auth_provider is None:
+            raise ValueError("agentic_identity requires an auth_provider")
+        auth_provider = self._auth_provider
+
+        return lambda: auth_provider.token(self._cloud.agentic_bot_scope, identity.tenant_id, identity)
```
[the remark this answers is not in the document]
heyitsaamir: this is fine. in the auth provider we should use the credential.tenantid if not provided

packages/api/src/microsoft_teams/api/clients/bot/token_client.py
```diff
@@ -144,3 +142,15 @@ async def get_graph(self, credentials: Credentials) -> GetBotTokenResponse:
...
+        token_provider = cast(Any, credentials.token)
+        parameters = list(inspect.signature(token_provider).parameters.values())
+        accepts_agentic_identity = any(
+            parameter.kind == inspect.Parameter.VAR_KEYWORD or parameter.name == "agentic_identity"
+            for parameter in parameters
+        )
+        if accepts_agentic_identity:
+            return cast(str | Awaitable[str], token_provider(scope, credentials.tenant_id, agentic_identity=None))
+
+        return cast(str | Awaitable[str], token_provider(scope, credentials.tenant_id))
```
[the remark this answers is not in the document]
heyitsaamir: agree

packages/api/src/microsoft_teams/api/auth/credentials.py
```diff
@@ -5,7 +5,13 @@
...
 from typing import Awaitable, Callable, Literal, Optional, Union
 
-from ..models import CustomBaseModel
+from ..models import AgenticIdentity, CustomBaseModel
+
+TokenProvider = Callable[
+    [Union[str, list[str]], Optional[str], Optional[AgenticIdentity]],
+    Union[str, Awaitable[str]],
+]
+LegacyTokenProvider = Callable[[Union[str, list[str]], Optional[str]], Union[str, Awaitable[str]]]
```
corinagum: Why not just make agentic_identity optional? Or fix the naming? This isn't really Legacy, right? Both scenarios will continue.

packages/apps/src/microsoft_teams/apps/options.py
```diff
@@ -8,14 +8,20 @@
...
 
 from .http.adapter import HttpServerAdapter
 from .plugins import PluginBase
 
+TokenProvider = Callable[
+    [Union[str, list[str]], Optional[str], Optional[AgenticIdentity]],
+    Union[str, Awaitable[str]],
+]
+LegacyTokenProvider = Callable[[Union[str, list[str]], Optional[str]], Union[str, Awaitable[str]]]
+
```
corinagum: These should be imported from api, not duplicated.

packages/api/src/microsoft_teams/api/auth/cloud_environment.py
```diff
@@ -35,6 +37,7 @@ class CloudEnvironment:
     login_endpoint="https://login.microsoftonline.com",
     login_tenant="botframework.com",
     bot_scope="https://api.botframework.com/.default",
+    agentic_bot_scope="https://botapi.skype.com/.default",
```
corinagum, line 40: Is the url truly the same for all cloud envs?

corinagum: Or just not available on all clouds yet? If yes, then maybe we should handle this differently? it will just silently fail for other cloud users

corinagum (commented):
Can agentic identity not create conversations or list members? Looks like only sending/modifying activities is performable?

packages/api/src/microsoft_teams/api/auth/credentials.py
heyitsaamir: Yeah, good question.

* In python if you pass in more values than a function accepts, it throws. So if people have a `def token(scope, tenant_id): str` function, and we pass in `token(scope, tenant_id, None)`, they'll all throw.
* In javascript, it's more forgiving, so we can probably do that, though people who are passing in two param tokens will just start to fail for agenticIdentities, unless they fix their token providers. (probably okay!)

heyitsaamir: Hmm, not totally sure what the more valid thing here is. I think keeping it Legacy is fine. Wdyt?

packages/apps/src/microsoft_teams/apps/options.py
heyitsaamir: Agree! Updated

packages/api/src/microsoft_teams/api/auth/cloud_environment.py
heyitsaamir: Asked the question to the folks that own this. let's see what they say. For now, I've included a TODO to update it when available

packages/api/src/microsoft_teams/api/clients/conversation/activity.py
```diff
@@ -35,10 +42,27 @@ def __init__(
...
+            http_client,
+            api_client_settings,
+            auth_provider=auth_provider,
+            agentic_identity=agentic_identity,
+            cloud=cloud,
+        )
         self.service_url = service_url.rstrip("/")
 
-    async def create(self, conversation_id: str, activity: ActivityParams) -> SentActivity:
+    def _get_service_url(self, service_url: str | None = None) -> str:
```
heyitsaamir: move this to the bottom.

packages/api/src/microsoft_teams/api/clients/conversation/client.py
```diff
@@ -83,11 +167,27 @@ def __init__(
...
+        self._activities_client = ConversationActivityClient(
+            self.service_url,
+            self.http,
+            self._api_client_settings,
+            auth_provider=auth_provider,
+            agentic_identity=agentic_identity,
+            cloud=cloud,
+        )
+        self._members_client = ConversationMemberClient(
+            self.service_url or "",
```
heyitsaamir: service url should be required right?

packages/api/src/microsoft_teams/api/clients/conversation/member.py
```diff
@@ -45,14 +62,19 @@ async def get(self, conversation_id: str) -> List[TeamsChannelAccount]:
         Returns:
             List of TeamsChannelAccount objects representing the conversation members
         """
-        response = await self.http.get(f"{self.service_url}/v3/conversations/{conversation_id}/members")
+        response = await self.http.get(
+            f"{self.service_url}/v3/conversations/{conversation_id}/members",
```
heyitsaamir: we should consolidate the service_url logic. we are doing two different things rn.

packages/api/src/microsoft_teams/api/clients/reaction/client.py
```diff
@@ -31,14 +39,22 @@ def __init__(
...
+        )
         self.service_url = service_url.rstrip("/")
 
     async def add(
         self,
         conversation_id: str,
         activity_id: str,
         reaction_type: MessageReactionType,
+        *,
+        agentic_identity: AgenticIdentity | None = None,
```
heyitsaamir, line 46: We should similarly add optional service url overrides here.

packages/api/src/microsoft_teams/api/clients/conversation/member.py
```diff
@@ -32,10 +41,18 @@ def __init__(
...
+            api_client_settings,
+            auth_provider=auth_provider,
+            agentic_identity=agentic_identity,
+            cloud=cloud,
+        )
         self.service_url = service_url.rstrip("/")
 
-    async def get(self, conversation_id: str) -> List[TeamsChannelAccount]:
+    async def get(
+        self, conversation_id: str, *, agentic_identity: AgenticIdentity | None = None
```
heyitsaamir: we should add similar service urls here.

packages/api/src/microsoft_teams/api/clients/meeting/client.py
```diff
@@ -30,10 +39,16 @@ def __init__(
...
+            options,
+            api_client_settings,
+            auth_provider=auth_provider,
+            agentic_identity=agentic_identity,
+            cloud=cloud,
+        )
         self.service_url = service_url.rstrip("/")
 
-    async def get_by_id(self, id: str) -> MeetingInfo:
+    async def get_by_id(self, id: str, *, agentic_identity: AgenticIdentity | None = None) -> MeetingInfo:
```
heyitsaamir: service_urls should be available as overrides for all these

packages/api/src/microsoft_teams/api/auth/credentials.py
corinagum: Yeah, that's fine. Could we just have a quick inline docs comment explaining the context of Legacy?

heyitsaamir: I renamed it to "Basic"

packages/api/src/microsoft_teams/api/clients/conversation/activity.py
```diff
@@ -146,14 +209,23 @@ async def create_targeted(self, conversation_id: str, activity: ActivityParams)
             The created activity
         """
         response = await self.http.post(
-            f"{self.service_url}/v3/conversations/{conversation_id}/activities?isTargetedActivity=true",
+            f"{self._get_service_url(service_url)}/v3/conversations/{conversation_id}/activities?isTargetedActivity=true",
             json=activity.model_dump(by_alias=True, exclude_none=True),
+            extensions={AGENTIC_IDENTITY_EXTENSION: self._resolve_agentic_identity(agentic_identity)},
```
heyitsaamir: targeted messages do not work with agentic identity

corinagum: I assume we're keeping agentic identity in the targeted messages methods for forward compat, right? Do we want to put this in inline docs in case customers try to use the method with agentic?

corinagum (approved):
lgtm! Just one nit

lilyydu (approved):
Looks pretty clean to me!

packages/api/src/microsoft_teams/api/clients/conversation/activity.py
heyitsaamir: Actually I just removed it.Need to follow up with the API team to figure out why it's not supported.

packages/api/src/microsoft_teams/api/clients/conversation/client.py
```diff
@@ -26,45 +27,128 @@ def __init__(self, client: "ConversationClient", conversation_id: str) -> None:
...
+        return await self._client.activities_client.get_members(
+            self._conversation_id, activity_id, service_url=service_url, agentic_identity=agentic_identity
+        )
 
-    async def create_targeted(self, activity: ActivityParams):
+    async def create_targeted(
+        self,
+        activity: ActivityParams,
+        *,
+        service_url: str | None = None,
```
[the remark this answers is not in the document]
heyitsaamir: yes intentional because targeted message operations don't support agentic identity yet

PR #479, opened 2026-06-17
Remove ActivitySender send abstraction

heyitsaamir:
Removes the separate `ActivitySender` class while preserving its send behavior.

Why:
ActivitySender had become a second send path. With AgenticIdentity support, the API client owns service URL resolution and request auth, so activity sends should route through the same API client path.

Interesting bits:
- `send_or_update_activity` keeps the old ActivitySender contract without keeping the class.
- Preserves create vs update behavior when `activity.id` is set.
- Preserves targeted create/update routing through `create_targeted` / `update_targeted`.
- Preserves `activity.from_ = ref.bot` and `activity.conversation = ref.conversation` stamping.
- Preserves targeted-in-personal-chat validation.
- ActivityContext still creates `HttpStream` directly from its scoped API client.

Reviewer tips:
Start with `activity_send.py`, then check `ActivityContext.send`, `App.send`, and `FunctionContext.send`.

Testing:
- Focused app send/reply/function/quoted-reply tests covering the old ActivitySender behavior.
- Ruff on touched app send files and tests.


heyitsaamir:

- #464
    - #479 ◀
        - #480

MehakBindra (approved):

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
```diff
@@ -191,7 +196,10 @@ async def send(
         self._add_targeted_message_info_entity(activity)
 
         ref = conversation_ref or self.conversation_ref
-        res = await self._activity_sender.send(activity, ref)
+        res = await self.api.conversations.activities(ref.conversation.id).create(
```
corinagum: Regression: `ActivitySender` routed to `create_targeted()` for targeted messages and `update()` > `updated_targted()` when `activity.id` is set. Without that routing, targeted messages will be sent as regular messages.

The tests that were deleted verified this routing, so those also need to be restored. 

`ActivitySender` also set `activity.from_ = ref.bot` and `activity.conversation = ref.conversation` before sending but not here. Those also should be restroed, as well as the tests.

corinagum (changes requested):
Removing ActivitySender is the right call, but it's introducing behavior regression between what people calling through it would experience vs. `api.conversation...create()` directly

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
heyitsaamir: Good catch!

packages/apps/src/microsoft_teams/apps/activity_send.py
```diff
@@ -0,0 +1,53 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from microsoft_teams.api import ActivityParams, AgenticIdentity, ApiClient, ConversationReference, MessageActivityInput
+from microsoft_teams.api.activities.sent_activity import SentActivity
```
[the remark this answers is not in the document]
heyitsaamir: agree!

heyitsaamir:
@corinagum the logic's updated again to move us to a utility that handles the complexities around sending. It's hared between ctx.send and app.send

packages/apps/src/microsoft_teams/apps/activity_send.py
```diff
@@ -0,0 +1,59 @@
...
+    SentActivity,
+)
+
+
+async def send_or_update_activity(
+    api: ApiClient,
+    activity: ActivityParams,
+    ref: ConversationReference,
+    *,
+    agentic_identity: AgenticIdentity | None = None,
```
lilyydu, line 21: is this a dead param?

corinagum: This is plumbing for the next PR :)

corinagum (approved):
LGTM!

lilyydu (approved):

PR #480, opened 2026-06-17
Add app-level agentic send helpers

heyitsaamir:
Adds app-level reactive and proactive helpers on top of the AgenticIdentity API surface.

Why:
The lower branches teach the API client how to send with `agentic_identity`. This branch wires that into the developer-facing app flows.

Reactive flow:
```python
@app.on_message
async def handle_message(ctx: ActivityContext[MessageActivity]):
    await ctx.reply(TypingActivityInput())

    if "react" in ctx.activity.text.lower():
        await ctx.api.reactions.add(
            conversation_id=ctx.activity.conversation.id,
            activity_id=ctx.activity.id,
            reaction_type="like",
        )
        await ctx.reply("Added a like reaction to your message.")
        return

    await ctx.send(f"You said '{ctx.activity.text}'")
```

Proactive app helper:
```python
agentic_identity = app.get_agentic_identity(agentic_app_id, agentic_user_id)

sent = await app.send(
    conversation_id,
    "Hello from app.send with an AgenticIdentity.",
    agentic_identity=agentic_identity,
)
```

Interesting bits:
- Reactive `ctx.api` is scoped with the inbound Agent ID identity when present.
- Reactive `ctx.send` / `ctx.reply` use the inbound service URL and identity through the API layer.
- `ActivityProcessor` now receives the app's existing auth provider instead of rebuilding one without credentials.
- Proactive `app.send(..., agentic_identity=...)` is a convenience over the same API operation path.

Reviewer tips:
Start with `examples/agent365/src/main.py`, then `ActivityProcessor` and `ActivityContext.send`.

Testing:
- Focused app helper, reactive context, and app-process auth-provider tests.
- Ruff on touched app/example files.

Live smoke tested with Agent ID token:
- reactive message send/reply flow
- proactive conversation activity create/update/reply/get_members
- reactions add/delete using canary service URL
- conversation members get_all/get/get_paged
- teams get_by_id/get_conversations
- conversations.create
- meetings get_by_id/get_participant

Known limitation:
- targeted activity create/update reached the service but returned 500. Leaving that as not proven until platform behavior is confirmed.

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/app.py
+++ b/packages/apps/src/microsoft_teams/apps/app.py
@@ -137,6 +137,7 @@ class App(ActivityHandlerMixin):
             self.options.default_connection_name,
             self.http_client,
             self._token_manager,
+            self._auth_provider,
             self.options.api_client_settings,
             self.cloud,
         )
--- a/packages/apps/src/microsoft_teams/apps/app.py
+++ b/packages/apps/src/microsoft_teams/apps/app.py
@@ -15,6 +15,7 @@ from microsoft_teams.api import (
     Account,
     ActivityBase,
     ActivityParams,
+    AgenticIdentity,
     ApiClient,
     ClientCredentials,
     ConversationAccount,
@@ -289,7 +290,14 @@ class App(ActivityHandlerMixin):
             self._events.emit("error", ErrorEvent(error, context={"method": "stop"}))
             raise
 
-    async def send(self, conversation_id: str, activity: str | ActivityParams | AdaptiveCard):
+    async def send(
+        self,
+        conversation_id: str,
+        activity: str | ActivityParams | AdaptiveCard,
+        *,
+        service_url: Optional[str] = None,
+        agentic_identity: Optional[AgenticIdentity] = None,
+    ) -> SentActivity:
         """Send an activity proactively to a conversation.
 
         Sends to the exact conversation ID provided. For channel threads,
@@ -305,7 +313,7 @@ class App(ActivityHandlerMixin):
 
         conversation_ref = ConversationReference(
             channel_id="msteams",
-            service_url=self.api.service_url,
+            service_url=service_url or self.api.service_url,
             bot=Account(id=self.id),
             conversation=ConversationAccount(id=conversation_id),
         )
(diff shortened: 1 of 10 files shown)


heyitsaamir:

- #464
    - #480 ◀

MehakBindra (approved):

packages/apps/src/microsoft_teams/apps/app.py
```diff
@@ -109,7 +109,7 @@ def __init__(self, **options: Unpack[AppOptions]):
 
         self.api = ApiClient(
             service_url,
-            self.http_client.clone(ClientOptions(token=self._get_bot_token)),
+            self.http_client,
```
corinagum: This removes the bot token from the HTTP client.  `_get_agentic_token`  returns `None` when there's no agentic identity, and nothing else provides a default token, so all non-agentic outbound API calls will be unauthenticated. Was this intentional?

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -90,8 +92,10 @@ async def _build_context(
         )
         api_client = ApiClient(
             service_url,
-            self.http_client.clone(ClientOptions(token=self.token_manager.get_bot_token)),
+            self.http_client,
             self.api_client_settings,
+            auth_provider=self.auth_provider,
+            agentic_identity=activity.recipient.agentic_identity,
```
corinagum, line 99: `recipient` is `Optional[Account]` & this will crash with `AttributeError` for any inbound activity without a recipient. Since `_build_context` runs for every activity, this could break the full flow.

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -56,6 +57,7 @@ def __init__(
         self.default_connection_name = default_connection_name
         self.http_client = http_client
         self.token_manager = token_manager
+        self.auth_provider = AppAuthProvider(token_manager)
```
corinagum: This omits credentials, as opposed to `App._init_`. `tenant_id` fallback in `AppAuthProvider.token()` won't work in reactive flows.

corinagum (changes requested):

packages/apps/src/microsoft_teams/apps/app.py
heyitsaamir: The auth provider should call get_app_token if agentic_identity is none:

https://github.com/microsoft/teams.py/blob/38321657777b9f6e963d9ae530bd930cd1d80f63/packages/apps/src/microsoft_teams/apps/auth_provider.py#L25

packages/apps/src/microsoft_teams/apps/app_process.py
heyitsaamir: Hmm recipient is actually `Account` on `ActivityBase` so it should be okay.

packages/apps/src/microsoft_teams/apps/app_process.py
heyitsaamir: `token_manager`, uses the tenant id from the credentials if not passed in. is that what you meant?

But maybe herer we can just pass app authprovider instead of reconstructing it with tokemanager.

corinagum: Makes sense to me

packages/apps/src/microsoft_teams/apps/app.py
corinagum: The problem is `auth_provider.token` is never called for non-agentic requests.

heyitsaamir: Ah shit. You're right. That's bad. Nice catch!

lilyydu (approved):
Nice!

corinagum (approved):
LGTM, thanks for addressing the feedback for all the Agentic Identity PRs

PR #484, opened 2026-06-22
Add host-owned ASGI route registration

### LINKED ISSUE #447: Support hosting the Teams endpoint inside a caller-owned ASGI app (sync route registration, host owns lifecycle)
## Summary

Please provide a **first-class, supported way to mount the Teams messaging endpoint onto an externally-owned ASGI app**, where the **host owns the server lifecycle** and route registration is available **synchronously** (decoupled from async plugin initialization).

Today the SDK is designed around the App owning (or at least driving) the HTTP server through an `HttpServerAdapter`. There is a "bring your own Starlette" pattern in `examples/http-adapters/src/starlette_adapter.py`, but it only works by calling the **async** `App.initialize()` to register routes, and the example adapter is not shipped in the package. That makes clean integration into a host that already owns the ASGI app, the port, and the startup/shutdown lifecycle surprisingly hard.

## Motivation / use case

We are building a multi-channel agent host (the Microsoft Agent Framework) where a single ASGI/Starlette application is owned by the host and serves many channels behind one public URL (Bot Service, Telegram, Discord, A2A, MCP, ...). Each channel *contributes* routes + startup/shutdown hooks; the host mounts them all and owns binding/serving.

To embed Teams there we want to:

1. Reuse the SDK's JWT validation, activity parsing, dispatch, and streaming (we don't want to reimplement Bot Framework auth).
2. **Not** let the SDK own the server — the host binds the port and runs uvicorn.
3. Obtain the registered routes **synchronously** at "contribution" time (before the event loop starts serving), because the host needs the route table up front, while async plugin `on_init` work must run later at startup.

## What we had to do today (and why it's painful)

Because route registration currently happens inside `HttpServer.initialize(...)` which is only invoked from the async `App.initialize()`, and because the only Starlette adapter is an example (not importable), we ended up:

- Writing a custom capturing `HttpServerAdapter` whose `register_route` just records `(method, path) -> handler`, and whose `serve_static` / `start` / `stop` are no-ops (host owns those).
- Calling the **private** `app.server.initialize(...)` ourselves to force synchronous route registration.
- Manually wiring `app.server.on_request = app._process_activity_event` (private) because that wiring normally happens inside `App.initialize()`, which we deliberately skip at route-collection time.
- Re-implementing the Starlette↔`HttpRequest`/`HttpResponse` translation, then calling the async `App.initialize()` later (at host startup) purely to run plugin `on_init` hooks — guarding against the SDK's private `_initialized` re-running the sync step.

This works but reaches into private internals (`server._adapter`, `server.initialize`, `_process_activity_event`, `_initialized`) and is fragile against minor SDK changes.

## Proposed API (illustrative)

A public, lifecycle-free integration seam. For example, any of:

- `app.register_routes() -> list[tuple[HttpMethod, str, HttpRouteHandler]]` — synchronously perform auth setup + route registration and return the route table, **without** owning a server and **without** running async plugin init.
- Split `App.initialize()` so that the synchronous "register routes / wire dispatcher" step is a separate, public, sync-callable method from the async plugin `on_init` step (e.g. `app.register_routes()` (sync) + `await app.start_plugins()` (async)).
- A **shipped** (not example-only) ASGI/Starlette adapter that supports "bring your own app, host owns lifecycle", with route registration callable without `await app.start()` and clearly separated from plugin init.

Concretely, the host wants something like:

```python
app = App(client_id=..., client_secret=..., messaging_endpoint="/teams/messages")
routes = app.register_routes()        # sync: auth + messaging endpoint, NO server, NO async plugin init
host.mount(routes)                    # host owns the ASGI app + port
# ... at host startup:
await app.start_plugins()             # async: plugin on_init only; never binds a port
```

## Acceptance criteria

- A documented, **non-private** way to obtain/register the Teams messaging route(s) onto a caller-owned ASGI app.
- Route registration callable **synchronously** (or at least separable from the async plugin-init step), so a host can build its route table before serving.
- The SDK never tries to bind a port / own start-stop in this mode.
- A shipped Starlette (and/or FastAPI) "bring your own app" adapter, so consumers don't copy the `examples/` adapter.

Happy to contribute a PR if the team is open to one of these shapes.

dcaayushd:
## Summary

Fixes #447.

Adds a public lifecycle split for host-owned ASGI integrations:

- Adds `App.register_routes()` to synchronously register the Teams messaging endpoint without starting a server or running async plugin init.
- Adds `App.start_plugins()` so hosts can run plugin `on_init` hooks later from their ASGI startup lifecycle.
- Returns a public `HttpRoute` route table from registration.
- Ships `StarletteAdapter` as part of `microsoft_teams.apps.http` instead of leaving it example-only.
- Updates host-owned FastAPI/Starlette examples and docs to use the new lifecycle.

## Validation

```bash
./.venv/bin/python -m pytest packages/apps/tests packages/botbuilder/tests
./.venv/bin/python -m ruff check packages/apps/src/microsoft_teams/apps packages/apps/tests/test_app.py packages/apps/tests/test_http_server.py examples/a2a/src/main.py examples/http-adapters/src/fastapi_non_managed.py examples/http-adapters/src/starlette_echo.py examples/mcp-server/src/main.py
./.venv/bin/pyright packages/apps/src/microsoft_teams/apps packages/botbuilder/src/microsoft_teams/botbuilder/adapter.py examples/a2a/src/main.py examples/http-adapters/src/fastapi_non_managed.py examples/http-adapters/src/starlette_echo.py examples/mcp-server/src/main.py

### DIFF
--- a/examples/a2a/src/main.py
+++ b/examples/a2a/src/main.py
@@ -80,6 +80,11 @@ async def handle_message(ctx: ActivityContext[MessageActivity]) -> None:
         await ctx.reply(reply)
 
 
+# Register Teams routes synchronously so the host-owned FastAPI route table is
+# complete before the server starts.
+app.register_routes()
+
+
 async def main() -> None:
     # Build this bot's own AgentCard — equivalent to AgentCardFactory.Build(config) in C#.
     agent_card = build_agent_card(config)
@@ -89,7 +94,7 @@ async def main() -> None:
     a2a_starlette = make_a2a_app(teams_app=app, agent=bot_agent, config=config, agent_card=agent_card)
     fastapi_app.mount("/a2a", a2a_starlette.build())
 
-    await app.initialize()
+    await app.start_plugins()
 
     port = int(getenv("PORT", "3978"))
     server = uvicorn.Server(uvicorn.Config(fastapi_app, host="0.0.0.0", port=port, log_level="info"))
--- a/examples/http-adapters/README.md
+++ b/examples/http-adapters/README.md
@@ -1,12 +1,12 @@
 # HTTP Adapters Examples
 
-Examples showing how to use custom `HttpServerAdapter` implementations and non-managed server patterns with the Teams Python SDK.
+Examples showing how to use shipped `HttpServerAdapter` implementations and non-managed server patterns with the Teams Python SDK.
 
 ## Examples
 
 ### 1. Starlette Adapter (`starlette_echo.py`)
 
-A custom `HttpServerAdapter` implementation for [Starlette](https://www.starlette.io/). Demonstrates how to write an adapter for any ASGI framework.
+The shipped `StarletteAdapter` implementation for [Starlette](https://www.starlette.io/).
 
 **Pattern**: Custom adapter, SDK-managed server lifecycle (`app.start()`)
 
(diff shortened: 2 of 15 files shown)


lilyydu:
Hey @dcaayushd ,

Thanks for jumping on this! This looks like we're headed in the right direction - though I don't have a strong pref towards shipping the StarletteAdapter as apart of apps - @heyitsaamir do you have any particular ops?

PR #485, opened 2026-06-22
Add agentic auth provider infrastructure

heyitsaamir:
Adds agentic auth provider plumbing to the API layer.

`ApiClient` is powered by `HttpClient` to make requests. Normally, `HttpClient` can take a token directly, but Agent ID calls need the SDK to choose between a regular bot token and an agentic identity token per request.

So this PR does the following:
1. Introduces the `AgenticIdentity` model.
2. Introduces an `AuthProvider` concept at the `ApiClient` layer. If provided, auth can happen through `ApiClient` if HttpClient does not have a token revolver. 
3. Installs an HTTP interceptor from `ApiClient`. The interceptor skips requests that already have `Authorization`, otherwise it uses the `AuthProvider`.
4. Uses `httpx` request extensions to mark per-request `AgenticIdentity` metadata. This is local-only metadata, not a header, so it does not leak to Teams endpoints. With this, we are basically stamping a request that requires agentic identity.
5. The interceptor reads that extension, picks the bot vs Agent ID scope, and passes the scope + identity to the `AuthProvider`.
6. Updates token-provider typing so custom token providers can receive `AgenticIdentity` when needed.
7. Sovereign cloud Agent ID scopes are TODO-marked until platform values are confirmed. No sneaky magic there.

### DIFF
--- a/examples/agent365/src/main.py
+++ b/examples/agent365/src/main.py
@@ -7,7 +7,7 @@ import asyncio
 import os
 
 from dotenv import load_dotenv
-from microsoft_teams.api import ClientCredentials
+from microsoft_teams.api import AgenticIdentity, ClientCredentials
 from microsoft_teams.apps.token_manager import AGENT_BOT_API_SCOPE, TokenManager
 
 
@@ -26,8 +26,7 @@ async def main():
     blueprint_client_id = get_required_env("AGENT365_BLUEPRINT_CLIENT_ID")
     blueprint_client_secret = get_required_env("AGENT365_BLUEPRINT_CLIENT_SECRET")
     agentic_app_id = get_required_env("AGENT365_AGENTIC_APP_ID")
-    agentic_user_id = os.getenv("AGENT365_AGENTIC_USER_ID")
-    agentic_user_upn = os.getenv("AGENT365_AGENTIC_USER_UPN")
+    agentic_user_id = get_required_env("AGENT365_AGENTIC_USER_ID")
     scope = os.getenv("AGENT365_SCOPE", AGENT_BOT_API_SCOPE)
 
     credentials = ClientCredentials(
@@ -38,11 +37,8 @@ async def main():
     token_manager = TokenManager(credentials=credentials)
 
     token = await token_manager.get_agentic_token(
-        tenant_id,
-        agentic_app_id,
         scope,
-        agentic_user_id=agentic_user_id,
-        agentic_user_upn=agentic_user_upn,
+        AgenticIdentity(agentic_app_id=agentic_app_id, agentic_user_id=agentic_user_id, tenant_id=tenant_id),
     )
 
     print(f"Acquired agent user token for {scope}")
--- a/packages/api/src/microsoft_teams/api/auth/credentials.py
+++ b/packages/api/src/microsoft_teams/api/auth/credentials.py
@@ -3,9 +3,34 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 Licensed under the MIT License.
 """
 
-from typing import Awaitable, Callable, Literal, Optional, Union
+from typing import Awaitable, Callable, Literal, Optional, Protocol, TypeAlias, Union, runtime_checkable
 
-from ..models import CustomBaseModel
+from ..models import AgenticIdentity, CustomBaseModel
+
+TokenScope: TypeAlias = Union[str, list[str]]
+TokenResult: TypeAlias = Union[str, Awaitable[str]]
+BasicTokenProvider: TypeAlias = Callable[[TokenScope, Optional[str]], TokenResult]
+_PositionalAgenticTokenProvider: TypeAlias = Callable[
+    [TokenScope, Optional[str], Optional[AgenticIdentity]], TokenResult
+]
+
+
+@runtime_checkable
+class _KeywordAgenticTokenProvider(Protocol):
+    def __call__(
+        self,
+        scope: TokenScope,
+        tenant_id: Optional[str],
+        *,
+        agentic_identity: Optional[AgenticIdentity] = None,
+    ) -> TokenResult: ...
+
+
+TokenProvider: TypeAlias = Union[
+    BasicTokenProvider,
+    _PositionalAgenticTokenProvider,
+    _KeywordAgenticTokenProvider,
+]
 
 
 class ClientCredentials(CustomBaseModel):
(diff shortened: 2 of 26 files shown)


heyitsaamir:

- #464
    - #485 ◀
        - #478
            - #479
                - #480

packages/api/src/microsoft_teams/api/clients/_auth_provider_interceptor.py
```diff
@@ -0,0 +1,31 @@
...
+        self._auth_provider = auth_provider
+        self._cloud = cloud or PUBLIC
+
+    async def request(self, ctx: InterceptorRequestContext) -> None:
+        if "Authorization" in ctx.request.headers:
+            return
+
+        token = await resolve_token(lambda: self._auth_provider.token(self._cloud.bot_scope))
+        if token is not None:
+            ctx.request.headers["Authorization"] = f"Bearer {token}"
```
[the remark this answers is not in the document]
heyitsaamir: Agree. Let's warn when this hapens and not produce the header.

packages/api/src/microsoft_teams/api/clients/base_client.py
```diff
@@ -42,3 +60,24 @@ def http(self) -> Client:
...
+        self._apply_auth_provider_interceptor()
+
+    def _apply_auth_provider_interceptor(self) -> None:
+        if self._auth_provider is None:
+            return
+
+        if any(isinstance(interceptor, AuthProviderInterceptor) for interceptor in self._http.interceptors):
+            return
+
+        self._http.use_interceptor(cast(Interceptor, AuthProviderInterceptor(self._auth_provider, cloud=self._cloud)))
```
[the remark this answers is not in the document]
heyitsaamir: This isn't worth it. We don't expect the clouds to change.

MehakBindra:
Q: Can the "agentic vs bot" decision live in one place (the TokenManager) rather than three?

Right now the choice of agentic token vs app/bot token - which is purely agentic_identity is None? - seems to be made in three spots that have to stay in sync:

BaseClient._get_auth_token picks agentic_bot_scope + passes the identity when one exists
AuthProviderInterceptor hardcodes bot_scope with no identity (the fallback/no-identity branch)
AppAuthProvider.token re-branches on agentic_identity is None to call get_app_token vs get_agentic_token
So the same boolean is computed in _get_auth_token, recomputed in AppAuthProvider, and the third case is baked into the interceptor — and scope selection (bot_scope vs agentic_bot_scope) is split between the API layer and the cloud config instead of sitting next to the minting logic.

Could the upstream layers just pass the identity (or None) and let TokenManager own the decision - i.e. a single get_token(scope_resource, identity) that internally chooses the bot vs agentic scope/tenant/mint path? The interceptor would then just be the identity=None case rather than a separate hardcoded branch.

heyitsaamir:
Hm, yeah it's a bit unclear.

Basically the way .net does this is when an api is called with agentic identity set to true, it sets a signal saying "hey this method will require an agentic identity token to be minted". And then before sending, the right token is minted in the authentication provider.

In python there is no equivalent way to say "hey this request will require an agentic identity token to be minted" later on. Instead of that, we pass in a token override if identity already exists. If there's no override, it falls back to the authprovider's basic bot-token minting (which uses the default bot-scope). 
(Maybe this can be cleaner and that would help simplify the design).

What you're suggesting is already being done by the auth provider. 
image

The token manager is a layer above, and just cares about token minting, so I think the authprovider is the right place to put this branching.

But you're right the token overriding is a bit messy. Let me see if I can remove the token-overriding in favor of just stamping a request with agentic identity instead.

heyitsaamir:
Update: @MehakBindra - check the latest update. Here we're doing something similar as .net. Methods with agentic identity would stamp the request using `extension` and a set key. (see https://github.com/microsoft/teams.py/pull/478 for eg). Then, the interceptor would check if "Authentication" headers were already set. If not, it would call the auth provider. I think the logic is pretty simple now. Thoughts?

examples/agent365/src/main.py
```diff
@@ -38,11 +37,8 @@ async def main():
     token_manager = TokenManager(credentials=credentials)
 
     token = await token_manager.get_agentic_token(
-        tenant_id,
-        agentic_app_id,
         scope,
-        agentic_user_id=agentic_user_id,
-        agentic_user_upn=agentic_user_upn,
+        AgenticIdentity(agentic_app_id=agentic_app_id, agentic_user_id=agentic_user_id, tenant_id=tenant_id),
```
lilyydu, line 41: we don't need the UPN anymore?

packages/api/src/microsoft_teams/api/auth/credentials.py
```diff
@@ -3,9 +3,34 @@
...
+        scope: TokenScope,
+        tenant_id: Optional[str],
+        *,
+        agentic_identity: Optional[AgenticIdentity] = None,
+    ) -> TokenResult: ...
+
+
+TokenProvider: TypeAlias = Union[
+    BasicTokenProvider,
+    _PositionalAgenticTokenProvider,
```
lilyydu, line 31: what's the diff betw Positional and Keyword?

packages/api/src/microsoft_teams/api/clients/bot/token_client.py
```diff
@@ -144,3 +142,30 @@ async def get_graph(self, credentials: Credentials) -> GetBotTokenResponse:
         )
 
         return GetBotTokenResponse.model_validate(res.json())
+
+    def _call_token_provider(self, credentials: TokenCredentials, scope: str) -> str | Awaitable[str]:
+        token_provider = cast(Any, credentials.token)
+        try:
+            parameters = list(inspect.signature(token_provider).parameters.values())
+        except (TypeError, ValueError):
+            return cast(str | Awaitable[str], token_provider(scope, credentials.tenant_id))
```
corinagum, line 151: Might be good to note here or on `token_manager` that these `_call_token_provider`s are intentionally kept separate.

corinagum (approved):

examples/agent365/src/main.py
heyitsaamir: Yeah, simplified it to only require agentic identity

packages/api/src/microsoft_teams/api/auth/credentials.py
heyitsaamir: say you have a fn like this:
```py
def greet(name, greeting):
    print(f"{greeting}, {name}")
```

You can call it like this:
`greet("Alice", "Hello")`
Here you're passing in params positionally.

Or you can call it like this:
`greet(greeting="Hello", name="Alice")`
Here order doesn't matter

`def f(pos_only, /, normal, *, kw_only)`

Here everything before / has to be positional, everything after / can be either, and everything after * must be keyword.

MehakBindra (approved):

packages/api/src/microsoft_teams/api/clients/bot/token_client.py
heyitsaamir: Yeah, honestly, I'm planning on deprecating / removing this client soon anyway, since it's not used anywhere

Issue #489, opened 2026-06-30
Support multiple response streams within a message handler


rkukura:
### Bug Description

If a message handler needs to stream part of the response, then show an `AdaptiveCard`, then stream more of the response, there is no supported SDK API to manage stream lifecycle when interleaving content types. Calling `ctx.stream.close()` and then `ctx._activity_sender.create_stream(ctx.conversation_ref)` to create a new stream works, but requires accessing the private `_activity_sender` property of `ctx`. A public method on `ctx` to create a new stream would avoid the need to access the private property.

### Steps to Reproduce

No error, but code potentially breaks if private details of the SDK change in a future version.

### Expected Behavior

A public API such as `ctx.new_stream()` would close the old stream if it hasn't already been closed and open a new stream.

### Actual Behavior

Application must call `ctx._activity_sender.create_stream(ctx.conversation_ref)` which involves accessing a private property.

### SDK Version

2.0.12

### Python Version

3.12.12

### Additional Context

This pattern is essential for conversational AI workflows where a bot needs to display structured data (e.g., an AdaptiveCard) to confirm a user selection or approve usage of a tool while streaming a long-form text response.

heyitsaamir:
This makes sense. I evaluated a number of options and I'm thinking we could do something like:
`ctx.stream.close(reset=True)` and re-use the stream. Let me put up a draft.

PR #495, opened 2026-07-02
Support resetting response streams

heyitsaamir:
## Summary

- make `HttpStream` reusable after close: a later `emit()` or `update()` starts a new streamed message cycle on the same stream instance
- keep repeated `close()` calls idempotent until another emit/update occurs
- update the stream example with a `multi-stream` flow that emits an Adaptive Card in the first stream final message, closes, then reuses `ctx.stream` for a second streamed segment
- add coverage for close idempotency followed by emit-driven stream reopening
- refresh `uv.lock` to include the missing `formatted-messaging` workspace member and dependency metadata

## Context

This addresses the scenario in #489 where a handler needs to:

1. stream part of a response,
2. include a normal message or card in the flow,
3. continue with another streamed response.

Before this change, the only way to do that was to close `ctx.stream` and then reach into the private `ctx._activity_sender.create_stream(...)` API to create another stream.

## Motivation

This is useful for conversational AI and tool-using bot flows where a response is not always one uninterrupted text stream.

For example, a bot may need to stream an explanation while it reasons or searches, include a structured checkpoint message, approval prompt, or card, and then continue streaming the final answer after that intermediate interaction.

Without a public lifecycle API, developers either have to avoid streaming for these richer flows or access private SDK internals to create a fresh stream. Reopening the stream on the next emit gives them a supported way to split one handler's response into multiple sequential streamed messages while keeping the SDK's one-stream-object model.

### `ctx.new_stream()`

We first considered adding a public `ctx.new_stream()` method that would create and return another stream object.

That solves the private API issue, but it makes the model feel like a handler can manage multiple stream objects at once. In practice, Teams appears to behave more reliably when there is only one active stream at a time. Multiple stream objects also complicate ownership, finalization, event handler registration, and retry behavior.

### Context-managed multiple streams

We also considered having `ActivityContext` track all streams created during a handler and close them all at the end.

That keeps cleanup centralized, but it still encourages multiple open streams during one handler. It also makes ordering less explicit: a stream could remain active while a normal message or a second stream is sent, which is exactly the behavior that seemed risky.

### `close(reset=True)`

We also tried an explicit `close(reset=True)` option. It worked, but it made the API more awkward and introduced a flag for what is conceptually the next stream cycle.

The approach in this PR keeps `close()` simple: it finalizes the current streamed message. Repeated `close()` calls remain idempotent. If the caller later emits or updates, the stream starts a new cycle automatically.

## Example

The stream example now includes a `multi-stream` path:

https://github.com/user-attachments/assets/91021174-fd9d-43c4-acb0-e38348ef6d12

```python
ctx.stream.emit("[stream 1] Starting the first streamed response.")
ctx.stream.emit(MessageActivityInput(text="...").add_card(card))
await ctx.stream.close()

ctx.stream.emit("[stream 2] Reusing ctx.stream after close().")
```

This demonstrates the intended lifecycle: one streamed message is finalized, and the same `ctx.stream` instance is reused for the next streamed message when the handler emits again.

Addresses #489

### DIFF
--- a/examples/stream/src/main.py
+++ b/examples/stream/src/main.py
@@ -31,10 +31,50 @@ STREAM_MESSAGES = [
     "✨ Stream test complete!",
 ]
 
+FIRST_STREAM_MESSAGES = [
+    "[stream 1] Starting the first streamed response. ",
+    "[stream 1] This is using the default ctx.stream instance. ",
+    "[stream 1] Next the handler will close with reset=True.",
+]
+
+SECOND_STREAM_MESSAGES = [
+    "[stream 2] Reusing ctx.stream after close(reset=True). ",
+    "[stream 2] This should render after the non-stream checkpoint message. ",
+    "[stream 2] The app processor will close this stream when the handler returns.",
+]
+
+
+def should_run_multi_stream(text: str | None) -> bool:
+    normalized = (text or "").lower().replace("-", " ")
+    return "multi stream" in normalized
+
 
 @app.on_message
 async def handle_message(ctx: ActivityContext[MessageActivity]):
     """Stream messages to the user on any message activity."""
+
+    if should_run_multi_stream(ctx.activity.text):
+        ctx.stream.update("Starting stream 1...")
+        await asyncio.sleep(1)
+
+        for message in FIRST_STREAM_MESSAGES:
+            await asyncio.sleep(0.5)
+            ctx.stream.emit(message)
+
+        await ctx.stream.close(reset=True)
+        await asyncio.sleep(1)
+
+        sent_message = await ctx.send("NON-STREAM MESSAGE BETWEEN STREAMS")
+        logger.info("Sent checkpoint message: %s", sent_message.id)
+        await asyncio.sleep(2)
+
+        ctx.stream.update("Starting stream 2...")
+        await asyncio.sleep(1)
+
+        for message in SECOND_STREAM_MESSAGES:
+            await asyncio.sleep(0.5)
+            ctx.stream.emit(message)
+        return
 
     ctx.stream.update("Stream starting...")
     await asyncio.sleep(1)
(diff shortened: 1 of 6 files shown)


examples/stream/src/main.py
```diff
@@ -31,11 +32,88 @@
...
+
+def should_send_simple_card(text: str | None) -> bool:
+    normalized = (text or "").lower().replace("-", " ")
+    return "simple card" in normalized
+
+
+def create_simple_card() -> AdaptiveCard:
+    return AdaptiveCard.model_validate(
+        {
+            "type": "AdaptiveCard",
```
lilyydu: nit: can we use the builder methods instead?

packages/apps/src/microsoft_teams/apps/http_stream.py
```diff
@@ -113,6 +113,10 @@ def emit(self, activity: Union[MessageActivityInput, TypingActivityInput, str])
         if self._canceled:
             raise StreamCancelledError("Stream has been cancelled.")
 
+        if self._result is not None:
+            logger.debug("starting a new streamed message after close")
+            self._result = None
```
lilyydu: should we call reset here?

heyitsaamir: Cleaned this up a bit so it reads better

packages/apps/tests/test_http_stream.py
```diff
@@ -453,3 +453,49 @@ async def test_close_waits_for_flush_to_complete(self, mock_api_client, conversa
...
+        with patcher:
+            stream = HttpStream(mock_api_client, conversation_reference)
+            stream.on_close(handle_close)
+
+            stream.emit("First streamed message")
+            await asyncio.sleep(0)
+            await self._run_scheduled_flushes(scheduled)
+
+            first_result = await stream.close()
+            await close_event.wait()
```
rajan-chari: This test will fail on a timeout of await if close_event is not triggered?

rajan-chari (approved):
LGTM.  My comment is a clarification question.

lilyydu (approved):

lilyydu (approved):

lilyydu (approved):

Issue #487, opened 2026-06-23
[Bug]: File share links not present in MessageActivity


Maxi-J:
### Bug Description

When the user attaches a file in teams - in the 1:1 chat with the Teams app - and that file is from some file share (SharePoint or OneDrive; like the file suggestions when clicking on the clip in Teams) and not a file from the local computer, there is nothing whatsoever in the MessageActivity context present. Not in ctx.attachment or otherwise. Therefore, these attachments are completely invisible to the Teams app. For local files all works as expected.

### Steps to Reproduce

1. Set up a message input hook
2. Attach a file from a file share like OneDrive or SharePoint that Teams probably automatically sends more as reference than as a file I'd assume
3. Check ctx.attachment and see there is nothing but the text message the user put in along with the attachment

### Expected Behavior

The attached file is present in ctx.attachment like with locally uploaded files.

### Actual Behavior

No trace of any attachment is present whatsoever.

### SDK Version

2.0.13.4

### Python Version

3.14.6

### Additional Context

The environment is an business setup with SSO for SharePoint, OneDrive, Teams etc and is connected together - in case that changes things.
If you need debug logs or something - feel free to let me know and how to get the information you need.

singhk97:
@Maxi-J thanks for opening this issue I've added this to the backlog. Will attempt to repro this scenario.

dcaayushd:
hey @Maxi-J , could you add a quick debug log and share the raw activity JSON? something like print(ctx.activity.model_dump_json(indent=2)) in your message handler. that'll show us exactly which field Teams is using for the SharePoint file so we can wire it up correctly. thanks!

Maxi-J:
Sure! I've ran the model dump for both the success case and the failure case.

<details>
<summary>Failure case</summary>

```json
{
  "serviceUrl": "https://smba.trafficmanager.net/emea/REDACTED/",
  "timestamp": "2026-07-03T13:44:04.349313Z",
  "locale": "de-DE",
  "localTimestamp": "2026-07-03T15:44:04.349313+02:00",
  "channelId": "msteams",
  "from": {
    "id": "REDACTED",
    "aadObjectId": "REDACTED",
    "type": null,
    "properties": null,
    "isTargeted": null,
    "name": "REDACTED"
  },
  "conversation": {
    "id": "REDACTED",
    "tenantId": "REDACTED",
    "conversationType": "personal",
    "name": null,
    "isGroup": null
  },
  "replyToId": null,
  "entities": [
    {
[...]

PR #498, opened 2026-07-08
feat: add shared_file_urls helper to extract SharePoint links from HTML attachments

dcaayushd:
Related to #487

when a user shares a file via the Teams compose box, Teams embeds the 
file link in a text/html attachment rather than sending structured file 
metadata. the activity payload has no file info, just HTML content.

added a shared_file_urls property on MessageActivity that parses 
SharePoint/OneDrive URLs from text/html attachments as a practical 
workaround for this Teams platform behavior.

added two tests: one confirming URL extraction works, one confirming 
empty list returned when no HTML attachments present.

### DIFF
--- a/packages/api/src/microsoft_teams/api/activities/message/message.py
+++ b/packages/api/src/microsoft_teams/api/activities/message/message.py
@@ -43,6 +43,24 @@ from ...models.entity import (
 from ..utils import StripMentionsTextOptions, strip_mentions_text
 
 
+def _is_sharepoint_url(url: str) -> bool:
+    hostname = urlparse(url).hostname
+    if hostname is None:
+        return False
+
+    host = hostname.lower()
+    return host == "sharepoint.com" or host.endswith(".sharepoint.com")
+
+
+def _sharepoint_urls_from_candidates(candidates: list[str]) -> list[str]:
+    urls: list[str] = []
+    for raw in candidates:
+        url: str = unescape(raw)
+        if _is_sharepoint_url(url):
+            urls.append(url)
+    return urls
+
+
 class _MessageBase(CustomBaseModel):
     """Base class containing shared message activity fields (all Optional except type)."""
 
@@ -152,7 +170,9 @@ class MessageActivity(_MessageBase, ActivityBase):
         """
         urls: list[str] = []
         for attachment in self.attachments or []:
-            if attachment.content_type != "text/html" or not attachment.content:
+            if attachment.content_type != "text/html":
+                continue
+            if not attachment.content:
                 continue
 
             html = str(attachment.content)
@@ -160,15 +180,7 @@ class MessageActivity(_MessageBase, ActivityBase):
             if not candidates:
                 candidates = re.findall(r"https?://[^\s\"<>]+", html)
 
-            for raw in candidates:
-                url: str = unescape(raw)
-                hostname = urlparse(url).hostname
-                if hostname is None:
-                    continue
-
-                host = hostname.lower()
-                if host == "sharepoint.com" or host.endswith(".sharepoint.com"):
-                    urls.append(url)
+            urls.extend(_sharepoint_urls_from_candidates(candidates))
 
         return urls
 
--- a/packages/api/src/microsoft_teams/api/activities/message/message.py
+++ b/packages/api/src/microsoft_teams/api/activities/message/message.py
@@ -147,15 +147,23 @@ class MessageActivity(_MessageBase, ActivityBase):
         structured file metadata. This property parses those URLs out.
         """
         import re
+        from urllib.parse import urlparse
 
-        urls = []
+        urls: list[str] = []
         for attachment in self.attachments or []:
-            if attachment.content_type == "text/html" and attachment.content:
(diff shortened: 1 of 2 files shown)


packages/api/src/microsoft_teams/api/activities/message/message.py
```diff
@@ -137,6 +137,35 @@ def strip_mentions_text(self, options: Optional[StripMentionsTextOptions] = None
...
+                continue
+
+            html = str(attachment.content)
+            candidates = re.findall(r"href=['\"](https?://[^'\"]+)['\"]", html, flags=re.IGNORECASE)
+            if not candidates:
+                candidates = re.findall(r"https?://[^\s\"<>]+", html)
+
+            for url in candidates:
+                host = urlparse(url).netloc.lower()
+                if host.endswith("sharepoint.com"):
```
lilyydu: nit: stronger security check with `if host == "sharepoint.com" or host.endswith(".sharepoint.com"):`

packages/api/src/microsoft_teams/api/activities/message/message.py
```diff
@@ -137,6 +137,35 @@ def strip_mentions_text(self, options: Optional[StripMentionsTextOptions] = None
             self.text = stripped_text
         return self
 
+    @property
+    def shared_file_urls(self) -> list[str]:
+        """
+        Extract SharePoint/OneDrive file URLs from HTML attachments.
```
lilyydu: does onedrive share the same host URL as well?

packages/api/src/microsoft_teams/api/activities/message/message.py
```diff
@@ -137,6 +137,35 @@ def strip_mentions_text(self, options: Optional[StripMentionsTextOptions] = None
...
+
+        When a user shares a file via the Teams compose box, Teams embeds
+        the file link inside a text/html attachment rather than sending
+        structured file metadata. This property parses those URLs out.
+        """
+        import re
```
lilyydu: nit: imports moved up to top of file

packages/api/src/microsoft_teams/api/activities/message/message.py
```diff
@@ -137,6 +137,35 @@ def strip_mentions_text(self, options: Optional[StripMentionsTextOptions] = None
...
+        from urllib.parse import urlparse
+
+        urls: list[str] = []
+        for attachment in self.attachments or []:
+            if attachment.content_type != "text/html" or not attachment.content:
```
lilyydu: can we also add a test case to hit this false case

packages/api/src/microsoft_teams/api/activities/message/message.py
```diff
@@ -137,6 +137,35 @@ def strip_mentions_text(self, options: Optional[StripMentionsTextOptions] = None
...
+                continue
+
+            html = str(attachment.content)
+            candidates = re.findall(r"href=['\"](https?://[^'\"]+)['\"]", html, flags=re.IGNORECASE)
+            if not candidates:
+                candidates = re.findall(r"https?://[^\s\"<>]+", html)
+
+            for url in candidates:
+                host = urlparse(url).netloc.lower()
```
lilyydu: nit: `url = unescape(raw)` before to handle malformed URLs

packages/api/src/microsoft_teams/api/activities/message/message.py
```diff
@@ -137,6 +137,35 @@ def strip_mentions_text(self, options: Optional[StripMentionsTextOptions] = None
...
+                if host.endswith("sharepoint.com"):
+                    urls.append(url)
```
lilyydu: not fully sure if this resolves the customer's failure case - this might be a fix on the backend service side. going to clarify on the issue!

dcaayushd: Thanks for reviewing. I addressed the code nits and added coverage for the attachment skip cases.

Changes made:
- Moved imports to the top of the file.
- Decode URLs with `unescape(raw)` before parsing.
- Tightened host validation to `host == "sharepoint.com" or host.endswith(".sharepoint.com")`.
- Clarified the docstring to refer to SharePoint-backed file URLs, including OneDrive for Business links that use SharePoint hosts.
- Added focused tests for:
  - exact SharePoint URL extraction,
  - OneDrive for Business SharePoint-hosted links,
  - non-HTML attachments being ignored,
  - HTML attachments without content being ignored,
  - non-SharePoint hosts being rejected.

I agree the customer’s full failure case may still need backend clarification, but this helper should safely cover the SDK-side workaround for file links embedded in `text/html` attachments.

lilyydu: The customer responded to the original issue and the failure case does require a change in our backend. The link should be returned in the `contentUrl` (same as success case). The PR assumes the link is embedded but its just a plain message text, so the regex will still return an empty array

We want to address the root issue directly so will close this and prioritize the fix accordingly. Thank you for being proactive with a fix, however!

lilyydu:
Closing as per discussion above! cc @corinagum

PR #500, opened 2026-07-08
fix(apps): specify specific string for 'Content stream is not allowed'

lilyydu:
## Problem

Our streaming error conditions checks strings for keywords - but two of the errors share "not allowed". The one we want to catch is 

- `Content stream is not allowed` — the streaming feature is disabled for the user/bot → **`StreamNotAllowedError`**

but it also caught:
- `Content stream is not allowed on an already completed streamed message` — the stream is already finalized → **terminal**, should be `TerminalStreamError`. 

This latter error isn't possible given our current implementation, so we don't need to specially handle this case


heyitsaamir (approved):
Not a huge fan of this string matching error checking. Let's chat with backend team to see if we can have another signal (or we can use exact string matching instead)

PR #515, opened 2026-07-20
Add OpenTelemetry app diagnostics

heyitsaamir:
Adds the Apps-side OpenTelemetry layer: inbound activity processing and handler telemetry.

## Why

This gives developers the other half of the story: Teams called the app, the SDK processed the activity, and the right handler ran. When paired with the API layer below it, the trace reads like: inbound activity → handler → outbound Teams API → auth. Much less spelunking in logs, thank goodness.

This PR intentionally leaves Agent365-specific integration out of the SDK telemetry pass because that will land separately.

## Interesting bits

- Adds `microsoft.teams.activity.process` around inbound activity processing.
- Adds `microsoft.teams.handler` around handler dispatch.
- Adds Apps metrics for activity processing and handler dispatch/duration.
- Uses `Microsoft.Teams.Apps` as the instrumentation scope, with the package version attached as the scope version.
- Restores Apps compatibility surfaces that the middleware refactor touched, including send/reply scoping and HTTP server compatibility.
- Removes the Agent365 baggage helper/export path from this stack.

## Reviewer tips

Start with the app processing path, then check handler telemetry and the compatibility follow-up. This PR stacks on the API layer, so API outbound spans live in the parent PR.

## Testing

Targeted Apps diagnostics/process tests pass after removing baggage. CI will rerun for this layer. The full stack was also live-smoked against Agent365/ACF and Echo/PokemonCatcher before splitting, with Aspire showing the expected Apps/API span nesting.


corinagum (changes requested):
Same feedback as TS

corinagum (approved):

PR #516, opened 2026-07-20
Agent365 OpenTelemetry: Agent365 baggage bridge

heyitsaamir:
## Summary

Adds the Agent365 OpenTelemetry baggage bridge on top of the token-provider surface in #530. This is the middle layer of a three-PR stack; the Microsoft OpenTelemetry dependency and runnable example remain in the top PR.

The bridge establishes Agent365 identity context before the SDK's root turn span, so Apps, API, auth, and host-created Agent365 spans share one context. It also provides a reusable scope opener for proactive work, where there is no inbound activity to derive identity from.

## Decisions

- **Enabled by default, explicitly disableable.** Inbound turns receive identifier-only Agent365 baggage unless `telemetry={"agent365": False}` is set. This matches the TypeScript SDK and keeps the enrichment at the single app-processing chokepoint.
- **Personal fields require explicit opt-in.** Names, emails, and descriptions are controlled independently through the include list; message content is never baggage. `user.id` remains in the default pseudonymous identifier set for cross-language parity.
- **Standard OpenTelemetry boundary.** SDK packages use only the standard OpenTelemetry API and configure no exporter or propagator. The host owns the OTel SDK and Microsoft distro.
- **Current Agent365 identity model.** Proactive scopes accept `AgenticUser`, matching the current Python SDK's `agentic_app_instance_id`, `agentic_user_id`, and blueprint terminology.

## Deferred

Exporter authentication, Microsoft OpenTelemetry configuration, reactive/proactive sample wiring, and the resulting lockfile change are isolated in the top example PR.

## Validation

The complete stack was validated live through Teams: the fresh Python trace appeared in Aspire with Apps/API/auth/`invoke_agent` nesting, and Defender produced Agent365 `InvokeAgent` rows. This isolated layer also passes its focused Apps test suite, Ruff, and Pyright.


corinagum (changes requested):
Same feedback as TS

lilyydu (approved):

corinagum (approved):

PR #530, opened 2026-07-30
Agent365 OpenTelemetry: SDK token-provider surface

heyitsaamir:
## Summary

This is the bottom of a three-PR Python Agent365 OpenTelemetry stack, aligned with [microsoft/teams.ts#680](https://github.com/microsoft/teams.ts/pull/680). It establishes the authentication seam later layers consume: named token-provider capabilities plus an Apps-owned `app.token_provider`, while preserving callable app-only credentials and the existing outbound auth span.

## Decisions

- **One token seam.** `ApiClient` now consumes the named provider directly; the API `AuthProvider` contract and Apps `AppAuthProvider` adapter are removed rather than retained as a second abstraction over the same grants.
- **Structural capabilities instead of signature inspection.** Runtime-checkable protocols model app-only, Agentic User, and Agentic App Instance grants independently, so providers can expose only the identities they can safely mint and consumers narrow with `isinstance`.
- **No identity fallback.** Agentic requests fail explicitly when a configured provider lacks the matching capability; silently substituting an app-only token would authenticate as the wrong identity.
- **Current Python identity model.** The implementation adapts the validated stack work to the current base's `AgenticUser` and `agentic_app_instance_id` terminology rather than carrying forward pre-rename names.

## Deferred

Agent365 baggage propagation/options, examples, and exporter/dependency work are intentionally excluded for the next two PRs; this layer adds no new runtime telemetry behavior.


packages/api/tests/unit/test_team_client.py
```diff
@@ -14,6 +14,14 @@
 from microsoft_teams.common.http import Client, ClientOptions
 
 
+class _TokenProviderAdapter:
+    def get_app_token(self, scope: str, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=None)
+
+    def get_agentic_user_token(self, scope: str, agentic_user: AgenticUser, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=agentic_user)
```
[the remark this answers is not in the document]
heyitsaamir: Fixed in 9704329b. The adapter now forwards the requested scope, and the assertions verify the public bot and Agentic User bot scopes.

packages/api/tests/unit/test_meeting_client.py
```diff
@@ -22,6 +22,14 @@
 from microsoft_teams.common.http import Client, ClientOptions
 
 
+class _TokenProviderAdapter:
+    def get_app_token(self, scope: str, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=None)
+
+    def get_agentic_user_token(self, scope: str, agentic_user: AgenticUser, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=agentic_user)
```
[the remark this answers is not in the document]
heyitsaamir: Fixed in 9704329b. The adapter now forwards the requested scope, and the tests assert the explicit bot or Agentic User scope.

packages/api/tests/unit/test_reaction_client.py
```diff
@@ -13,6 +13,14 @@
 from microsoft_teams.api.models import AgenticUser
 
 
+class _TokenProviderAdapter:
+    def get_app_token(self, scope: str, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=None)
+
+    def get_agentic_user_token(self, scope: str, agentic_user: AgenticUser, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=agentic_user)
```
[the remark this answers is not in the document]
heyitsaamir: Fixed in 9704329b. Reaction tests now preserve and assert the explicit scope, including the overridden Agentic User scope.

packages/api/tests/unit/test_conversation_client.py
```diff
@@ -21,6 +21,14 @@
 from opentelemetry.trace import Span, SpanKind
 
 
+class _TokenProviderAdapter:
+    def get_app_token(self, scope: str, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=None)
+
+    def get_agentic_user_token(self, scope: str, agentic_user: AgenticUser, tenant_id: str | None = None):
+        return self.token(scope=None, agentic_user=agentic_user)
```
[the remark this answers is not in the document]
heyitsaamir: Fixed in 9704329b. Conversation tests now forward and assert the actual bot/Agentic User scopes instead of discarding them.

packages/apps/src/microsoft_teams/apps/token_provider.py
```diff
@@ -0,0 +1,53 @@
...
+        self,
+        scope: str | None,
+        agentic_user: AgenticUser,
+        tenant_id: str | None = None,
+    ) -> TokenProtocol | None:
+        """Acquire a token carrying an Agentic User identity."""
+        return await self._token_manager.get_agentic_user_token(
+            scope or self._cloud.agent_bot_scope,
+            agentic_user,
+        )
```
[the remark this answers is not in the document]
heyitsaamir: Aligned with teams.ts in 9704329b: the public AppTokenProvider method and API-client factory use the two-argument Agentic User method. The optional tenant remains only on the consumer-supplied provider protocol, where TokenManager passes its resolved tenant.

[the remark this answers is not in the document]
heyitsaamir: Follow-up in 28cc8965: the Python capability protocol now also uses the two-argument Agentic User method so AppTokenProvider statically satisfies every protocol. TokenManager carries its resolved tenant on AgenticUser before invoking custom providers.

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -149,7 +147,10 @@ async def _build_context(
             conversation_ref,
             is_signed_in,
             self.default_connection_name,
-            app_token=lambda: self.token_manager.get_graph_token(tenant_id),
+            app_token=lambda: self.token_provider.get_app_token(
```
lilyydu: if the incoming activity doesn't have a tenant ID, this sets it as `"common"` instead of falling back to the app tenant. before, `get_graph_token(None)` would try `credentials.tenant_id` first

`app.py:664` still does the credential fallback, so the same token is resolved 2 diff ways depending on which path gets hit

should the fallback move back inside `TokenManager` so there's only one version of this logic?

packages/api/src/microsoft_teams/api/clients/bot/token_client.py
```diff
@@ -143,19 +139,29 @@ async def get_graph(self, credentials: Credentials) -> GetBotTokenResponse:
...
             parameters = list(inspect.signature(token_provider).parameters.values())
         except (TypeError, ValueError):
-            return cast(str | Awaitable[str], token_provider(scope, credentials.tenant_id))
+            return cast(TokenResult, token_provider(scope, credentials.tenant_id))
 
         accepts_agentic_user = any(
             parameter.kind == inspect.Parameter.VAR_KEYWORD or parameter.name == "agentic_user"
             for parameter in parameters
         )
         if accepts_agentic_user:
```
lilyydu: bc we updated the union, I think this code is dead now, same w the below!

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -112,7 +110,7 @@ async def _build_context(
             service_url,
             self.http_client,
```
corinagum, line 112: Please pass `cloud=self.cloud` to this. Otherwise, per-activity clients in sovereign clouds request the wrong Bot Framework token scope

corinagum: Does this need updating in TS?

packages/api/src/microsoft_teams/api/clients/bot/token_client.py
heyitsaamir: Fixed in bedbcf21. The callable branch now invokes exactly `(scope, tenant_id)` per the updated union; the signature introspection, casts, and obsolete three-argument compatibility tests are removed.

packages/apps/src/microsoft_teams/apps/app_process.py
heyitsaamir: Fixed in bedbcf21. ActivityProcessor now passes the raw activity tenant to the app-owned Graph token callback, so the single App path preserves explicit tenant -> credential tenant -> `common`. The regression test invokes the callback and verifies the credential fallback.

packages/apps/src/microsoft_teams/apps/app_process.py
heyitsaamir: Fixed in bedbcf21 with sovereign-cloud regression coverage. No TS change is needed for this specific issue: teams.ts clones the app-owned API client, which retains its configured cloud; Python constructs a fresh per-activity ApiClient and therefore must pass `cloud=self.cloud` explicitly.

corinagum (approved):

packages/apps/src/microsoft_teams/apps/app_process.py
heyitsaamir: Looks like this was an issue even on main...

PR #531, opened 2026-07-30
Wire the Agent365 example onto the OpenTelemetry SDK surface

heyitsaamir:
## Summary

This is the top layer of the Python Agent365 OpenTelemetry stack, building on the automatic baggage bridge in #516 and the token-provider surface in #530. It makes those SDK seams runnable in the Agent365 example and mirrors the app-owned integration demonstrated by [microsoft/teams.ts#679](https://github.com/microsoft/teams.ts/pull/679).

## Decisions

- **Microsoft OpenTelemetry remains app-owned.** `microsoft-opentelemetry` is declared only by `examples/agent365`; no SDK package imports it or acquires an exporter dependency. The lockfile change is comparatively large because Microsoft OTel 1.3.6 brings its supported Azure Monitor, OTLP, web-framework, HTTP, logging, and OpenAI instrumentation set transitively into the example environment.
- **Async token acquisition is bridged to the synchronous exporter contract.** The example refreshes an Agentic App Instance token through `app.token_provider` before opening each Agent365 operation scope, then serves it from a synchronous in-memory resolver. This keeps credential ownership on the existing SDK token surface without blocking inside the exporter callback.
- **Correlation does not enable transcript capture.** Sensitive-data export remains disabled, message bodies are not logged, and `InvokeAgentScope` is created without input or output content. The unused `openai_agents` instrumentation is explicitly disabled rather than adding an Agents SDK dependency.
- **Short-lived proactive sends flush explicitly.** The proactive entry point force-flushes after its scoped work so the batching exporter does not lose the final operation when the process exits.

## Deferred

SDK token and baggage behavior stays in the lower stack layers (#530 and #516); this PR intentionally contains only the runnable example, its app-level dependency/configuration, documentation, and generated lockfile.


examples/agent365/src/main.py
```diff
@@ -128,22 +142,52 @@ async def handle_agentic_user_workload_onboarding_updated(
...
+        ctx.activity.recipient.id,
+    )
+
+    agentic_user = ctx.activity.recipient.agentic_user
+    if agentic_user is None or agentic_user.tenant_id is None:
+        logger.warning("No Agent365 user on the activity; handling without an InvokeAgent scope")
+        await _handle_message(ctx)
+        return
 
+    await token_cache.refresh(
```
corinagum, line 163: this runs before `_handle_message`, so should failed oauth be logged so message handling can continue without telemetry?

corinagum (approved):

examples/agent365/src/main.py
heyitsaamir: nah - the app wasn't actually able to do anything. this is an sdk level failure, so OTEL captures it.

lilyydu (approved):

PR #550, opened 2026-08-06
Feature: Receive files in personal scope

corinagum:
Adds API for receiving files to the app in personal scope. A user uploads a file, which arrives on the activity, and `ctx.files.list()` provides metadata of the file(s) and lazy handlers to download it. The url is `tempauth` and will expire.

- **`ctx.files`**: `list()` / `first()` map inbound `file.download.info` attachments to `IncomingFile`. Never throws: non-message activities and empty/malformed attachments yield an empty list; bad entries are skipped.
- **`IncomingFile`**: file metadata with lazy `download()`, `stream()`, `text(encoding=...)`, `save_as(path)`. Each read re-fetches (bounded by the short-lived URL).
- **`DownloadedFile`**: buffered snapshot you own (`bytes`, `content_type`, `filename`, `source_url`), with synchronous `text()`/`save_as()`.
- **Errors**: Differentiates between first fetch error (url expired before first fetch), reread error (url expired after first fetch), and scope not supported error (today, only 1:1).
- **Types**: new `FileDownloadInfo` model + `file.download.info` content-type constant; `ConversationType` extended with `channel` (forward-compatible).

## Testing

- Unit tests: `test_files_accessor.py`, `test_incoming_file.py`
- Manual e2e

## Note

- After investigating thumbnail url behavior, I adjusted `Attachments.thumbnail_url` docstrings to indicate it will not be populated on incoming files.
- Similarly for `etag`, this field may be set by Graph, which is not in scope for this PR.

### DIFF
--- a/packages/api/src/microsoft_teams/api/models/attachment/attachment.py
+++ b/packages/api/src/microsoft_teams/api/models/attachment/attachment.py
@@ -24,4 +24,7 @@ class Attachment(CustomBaseModel):
     "The name of the attachment"
 
     thumbnail_url: Optional[str] = None
-    "Thumbnail associated with attachment"
+    """
+    Thumbnail associated with attachment.
+    Not set by Teams when a bot receives an upload.
+    """
--- a/packages/api/src/microsoft_teams/api/models/conversation_type.py
+++ b/packages/api/src/microsoft_teams/api/models/conversation_type.py
@@ -1,0 +1,12 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from typing import Literal, Union
+
+# Define the literal types for known conversation types
+KnownConversationType = Literal["personal", "groupChat", "channel"]
+
+# Type alias for conversation type that can be either a known type or any other string
+ConversationType = Union[KnownConversationType, str]
--- a/packages/api/src/microsoft_teams/api/models/file/file_download_info.py
+++ b/packages/api/src/microsoft_teams/api/models/file/file_download_info.py
@@ -1,0 +1,33 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from typing import Any, Optional
+
+from ..custom_base_model import CustomBaseModel
+
+FILE_DOWNLOAD_INFO_CONTENT_TYPE = "application/vnd.microsoft.teams.file.download.info"
+"""
+Content type of an inbound uploaded-file attachment.
+Its `content` is a `FileDownloadInfo` describing a file fetchable from a short-lived, pre-authorized download URL.
+"""
+
+
+class FileDownloadInfo(CustomBaseModel):
+    """
+    The content of a `file.download.info` attachment, describing an uploaded file received in a personal (1:1) chat.
+    The file is fetched from the short-lived, pre-authorized `download_url` with a plain GET (no bearer token).
+    """
+
+    download_url: Optional[str] = None
+    "Pre-authorized, short-lived URL the file can be fetched from with a plain GET (no bearer token)."
+
+    unique_id: Optional[str] = None
+    "The OneDrive/ODSP drive-item id for the file. This is the storage-specific file identity a Graph fetch keys off."
+
+    file_type: Optional[str] = None
+    "Type of file (extension, e.g. `pdf`, `docx`)."
+
+    etag: Optional[Any] = None
+    "ETag for the file."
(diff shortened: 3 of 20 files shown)


packages/api/src/microsoft_teams/api/models/conversation_type.py
```diff
@@ -0,0 +1,12 @@
...
+Licensed under the MIT License.
+"""
+
+from typing import Literal, Union
+
+# Define the literal types for known conversation types
+KnownConversationType = Literal["personal", "groupChat", "channel"]
+
+# Type alias for conversation type that can be either a known type or any other string
+ConversationType = Union[KnownConversationType, str]
```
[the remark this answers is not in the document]
corinagum: intentional for forward compat

packages/api/src/microsoft_teams/api/models/attachment/attachment.py
```diff
@@ -24,4 +24,7 @@ class Attachment(CustomBaseModel):
     "The name of the attachment"
 
     thumbnail_url: Optional[str] = None
-    "Thumbnail associated with attachment"
+    """
+    Thumbnail associated with attachment.
+    Not set by Teams when a bot receives an upload.
```
lilyydu, line 29: qq: does this get set by the client later downstream?

packages/api/src/microsoft_teams/api/models/file/file_download_info.py
```diff
@@ -0,0 +1,36 @@
...
+    The file is fetched from the short-lived, pre-authorized `download_url` with a plain GET (no bearer token).
+    """
+
+    download_url: Optional[str] = None
+    "Pre-authorized, short-lived URL the file can be fetched from with a plain GET (no bearer token)."
+
+    unique_id: Optional[str] = None
+    "The OneDrive/ODSP drive-item id for the file. This is the storage-specific file identity a Graph fetch keys off."
+
+    file_type: Optional[str] = None
```
lilyydu, line 29: should this be a required field?

packages/apps/src/microsoft_teams/apps/files/files_accessor.py
```diff
@@ -0,0 +1,130 @@
...
+    only through the raw `activity.attachments` array.
+
+    This covers the file-upload path, not "any uploaded media". What matters is how the content arrived, not the
+    file's MIME type, so file *type* is unrestricted (pdf, docx, png, etc.) as long as it was sent as an uploaded
+    file. An image sent as a file appears here, but the same image pasted inline does not.
+    """
+
+    def __init__(self, activity: ActivityBase, logger: logging.Logger) -> None:
+        self._activity = activity
+        self._logger = logger
```
lilyydu: I think you should be able to do `logger = logging.getLogger(__name__)` at top level without passing it in

packages/apps/src/microsoft_teams/apps/files/files_accessor.py
```diff
@@ -0,0 +1,130 @@
...
+        resolution so it stays correct when later scopes hydrate through Graph.
+        """
+        files = await self.list()
+        return files[0] if files else None
+
+    def _detect_scope(self) -> ConversationType:
+        """Derive the conversation scope from the inbound activity."""
+        conversation = getattr(self._activity, "conversation", None)
+        conversation_type = getattr(conversation, "conversation_type", None)
+        return conversation_type or "personal"
```
lilyydu, line 82: should we be defaulting to "personal" here?

packages/api/src/microsoft_teams/api/models/attachment/attachment.py
corinagum: No, only on outbound files (send)

packages/api/src/microsoft_teams/api/models/file/file_download_info.py
corinagum: No, that would silently drop files. Teams sometimes omits `fileType`, so I added a test for when it is `None`. If it were required, it would raise on `model_validate` and skip the attachment, causing it to disappear from `list()`. Matches `FileInfoCard` and `FileUploadInfo`

packages/apps/src/microsoft_teams/apps/files/files_accessor.py
corinagum: Did this deliberately, otherwise it would entail making `scope` optional, which adds an extra branch of validation for the download and error mapping. But defaulting to 'personal' would only apply if `conversation.conversation_type` is missing, which Teams always populates.

lilyydu (approved):

PR #553, opened 2026-08-10
feat: add per-turn state foundation (state scopes, container, loader)

lilyydu:
This is **PR 1 of 5** in the multi-connection OAuth stack. It introduces the per-turn state foundation independently so later PRs can add `ctx.state`, OAuth routing, and token-exchange deduplication.

This PR does **not** wire state into activity dispatch. That follows in PR 2.

### `TurnState`

A Pythonic `MutableMapping[str, Any]` representing one conversation or user state scope.

- Supports normal mapping operations such as `state["key"]`, assignment, and deletion.
- Detects nested `dict` and `list` mutations using content snapshots.
- Avoids writes when content is unchanged or a mutation is reverted.
- Seals after the turn so retained state cannot be accessed later.
- Requires values to be JSON-serializable.

### `TurnStateContainer`

Bundles the conversation scope and optional user scope with the identity they were loaded for.

- Uses keyword-only fields.
- Saves using the original conversation and user IDs.
- Deletes backing storage before clearing in-memory state.
- Marks cleared scopes clean so end-of-turn persistence does not repeat the deletion.

### `TurnStateLoader`

Loads and persists scopes through the existing `Storage` abstraction.

- Uses `{prefix}:conv:{conversationId}` and `{prefix}:user:{conversationId}:{userId}` keys.
- Percent-encodes dynamic key segments to prevent ambiguous or colliding keys.
- Persists each scope as a raw JSON object string.
- Skips clean scopes and deletes dirty scopes that become empty.
- Serializes all dirty scopes before performing storage writes, preventing serialization failures from partially updating state.
- Marks scopes clean only after all requested storage operations succeed.
- Treats unreadable or malformed values as missing and removes them from storage.
- Accepts already-deserialized dictionaries for compatibility with storage providers.

### Provider-owned storage behavior

The state layer keeps the existing `Storage.get()` / `set()` / `delete()` contract unchanged.

- Expiration and other storage-specific behavior are configured directly on the provider supplied through `StateOptions.storage`.
- State has no expiration by default.
- Applications can provide a dedicated, preconfigured state storage instance when they need provider-specific policies.

## Cross-SDK alignment

The implementation follows the same behavioral model as Teams SDK state in C# and TypeScript:

- Conversation and user scopes.
- Load once and save dirty scopes at the end of a turn.
- Empty dirty scopes delete their backing keys.
- State becomes unavailable after the turn.
- Raw JSON scope persistence.
- No expiration unless the configured provider applies it.

The public API remains Pythonic: `TurnState` behaves like a mapping rather than copying TypeScript's explicit `get()`/`set()` API. Python also automatically detects nested mutations so callers do not need to remember to reassign mutable values.

## Validation

- 37 focused state tests.
- Coverage includes nested mutations, reverted changes, sealing, deletion, malformed data, escaped keys, save ordering, and successful-save re-baselining.
- Build, lint, type-check, and supported Python-version checks pass.

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/state/container.py
+++ b/packages/apps/src/microsoft_teams/apps/state/container.py
@@ -44,11 +44,13 @@ class TurnStateContainer:
     async def delete(self) -> None:
         """Clear both scopes and remove them from the backing store.
 
-        Clearing marks the scopes dirty+empty (so a later save deletes the keys), and the
-        injected deleter removes the keys immediately.
+        The injected deleter removes the keys immediately, then in-memory scopes
+        are cleared so state reflects the deletion during the current turn.
         """
+        if self._deleter is None:
+            raise RuntimeError("State deletion is not available. Call UseState() during service registration.")
+
+        await self._deleter()
         self.conversation.clear()
         if self.user is not None:
             self.user.clear()
-        if self._deleter is not None:
-            await self._deleter()
--- a/packages/apps/src/microsoft_teams/apps/state/container.py
+++ b/packages/apps/src/microsoft_teams/apps/state/container.py
@@ -1,0 +1,51 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from typing import Awaitable, Callable, Optional
+
+from .turn_state import TurnState
+
+_Deleter = Callable[[], Awaitable[None]]
+
+
+@dataclass
+class TurnStateContainer:
+    """The state scopes loaded for one turn, together with the identity they
+    were loaded for.
+
+    ``conversation`` is always present. ``user`` is ``None`` when the activity has
+    no ``from`` identity, so there is no per-user scope to load or persist.
+
+    ``conversation_id``/``user_id`` record the identity this container was loaded
+    for. The loader reads them back off the container when saving, so a save can
+    never be told to persist under a different key than it was loaded from.
+    """
+
+    conversation: TurnState
+    user: Optional[TurnState] = None
+    conversation_id: str = ""
+    user_id: Optional[str] = None
+    _deleter: Optional[_Deleter] = field(default=None, repr=False, compare=False)
+
+    def seal(self) -> None:
+        """Seal every scope so post-turn access raises."""
+        self.conversation.seal()
+        if self.user is not None:
+            self.user.seal()
+
+    async def delete(self) -> None:
+        """Clear both scopes and remove them from the backing store.
+
+        Clearing marks the scopes empty so a later save is a no-op, and the
+        injected deleter removes the keys immediately.
+        """
+        self.conversation.clear()
+        if self.user is not None:
+            self.user.clear()
+        if self._deleter is not None:
+            await self._deleter()
--- a/packages/apps/src/microsoft_teams/apps/state/container.py
+++ b/packages/apps/src/microsoft_teams/apps/state/container.py
@@ -41,7 +41,7 @@ class TurnStateContainer:
     async def delete(self) -> None:
         """Clear both scopes and remove them from the backing store.
 
-        Clearing marks the scopes empty so a later save is a no-op, and the
+        Clearing marks the scopes dirty+empty (so a later save deletes the keys), and the
         injected deleter removes the keys immediately.
         """
         self.conversation.clear()
--- a/packages/apps/src/microsoft_teams/apps/state/container.py
+++ b/packages/apps/src/microsoft_teams/apps/state/container.py
@@ -13,7 +13,7 @@ from .turn_state import TurnState
 _Deleter = Callable[[], Awaitable[None]]
 
 
-@dataclass
+@dataclass(kw_only=True)
 class TurnStateContainer:
     """The state scopes loaded for one turn, together with the identity they
     were loaded for.
@@ -24,11 +24,14 @@ class TurnStateContainer:
     ``conversation_id``/``user_id`` record the identity this container was loaded
     for. The loader reads them back off the container when saving, so a save can
     never be told to persist under a different key than it was loaded from.
+
+    Fields are keyword-only so the public constructor is not tied to positional
+    order and can evolve without breaking callers.
     """
 
     conversation: TurnState
+    conversation_id: str
     user: Optional[TurnState] = None
-    conversation_id: str = ""
     user_id: Optional[str] = None
     _deleter: Optional[_Deleter] = field(default=None, repr=False, compare=False)
 
--- a/packages/apps/src/microsoft_teams/apps/state/container.py
+++ b/packages/apps/src/microsoft_teams/apps/state/container.py
@@ -52,5 +52,7 @@ class TurnStateContainer:
 
         await self._deleter()
         self.conversation.clear()
+        self.conversation.mark_clean()
         if self.user is not None:
             self.user.clear()
+            self.user.mark_clean()
--- a/packages/apps/src/microsoft_teams/apps/state/loader.py
+++ b/packages/apps/src/microsoft_teams/apps/state/loader.py
@@ -7,7 +7,6 @@ from __future__ import annotations
 
 import json
 import logging
-import time
 from typing import Any, Dict, Optional, cast
 from urllib.parse import quote
 
@@ -24,17 +23,16 @@ class TurnStateLoader:
     """Loads and persists :class:`TurnState` scopes over a ``Storage`` backend.
 
     Values are stored as JSON **strings** so any ``Storage`` implementation works
-    regardless of how it serializes values. Each blob carries a save timestamp
-    that powers the loader-applied TTL, since ``Storage`` has no native expiry.
-
+    regardless of how it serializes values. Expiry and other write behavior are
+    delegated to the storage implementation through ``StorageOptions``.
     """
 
-    def __init__(self, storage: Optional[Storage[str, str]] = None, options: Optional[StateOptions] = None) -> None:
+    def __init__(self, storage: Optional[Storage[str, Any]] = None, options: Optional[StateOptions] = None) -> None:
         self._options = options or StateOptions()
         resolved = storage if storage is not None else self._options.storage
         if resolved is None:
             raise ValueError("TurnStateLoader requires a Storage backend (pass one explicitly or via StateOptions).")
-        self._storage: Storage[str, str] = resolved
+        self._storage: Storage[str, Any] = resolved
 
     @property
     def options(self) -> StateOptions:
@@ -101,7 +99,10 @@ class TurnStateLoader:
         for key in pending_deletes:
             await self._storage.async_delete(key)
         for key, value in pending_sets:
-            await self._storage.async_set(key, value)
+            if self._options.storage_options is None:
+                await self._storage.async_set(key, value)
+            else:
+                await self._storage.async_set_with_options(key, value, self._options.storage_options)
         for scope in pending_clean:
             scope.mark_clean()
 
@@ -116,17 +117,12 @@ class TurnStateLoader:
         if raw is None:
             return TurnState()
 
-        blob = self._deserialize(raw)
-        if blob is None or self._is_expired(blob):
+        data = self._deserialize(raw)
+        if data is None:
             await self._storage.async_delete(key)
             return TurnState()
 
-        data = blob.get("data")
-        if not isinstance(data, dict):
-            await self._storage.async_delete(key)
-            return TurnState()
-
-        return TurnState(cast(Dict[str, Any], data))
+        return TurnState(data)
 
     def _prepare_scope_save(
         self,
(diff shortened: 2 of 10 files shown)


packages/apps/src/microsoft_teams/apps/state/container.py
```diff
@@ -0,0 +1,51 @@
...
+    no ``from`` identity, so there is no per-user scope to load or persist.
+
+    ``conversation_id``/``user_id`` record the identity this container was loaded
+    for. The loader reads them back off the container when saving, so a save can
+    never be told to persist under a different key than it was loaded from.
+    """
+
+    conversation: TurnState
+    user: Optional[TurnState] = None
+    conversation_id: str = ""
```
corinagum: #554 brings `TurnStateContainer` to public API surface, so the positional fields will need to be frozen. maybe update the order? ` conversation, conversation_id*, user=None, user_id=None`

`@dataclass(kw_only=True)` ? 

*Are we sure we want empty string as default for conversation_id?

packages/apps/src/microsoft_teams/apps/state/turn_state.py
```diff
@@ -0,0 +1,97 @@
...
+        self._sealed = True
+
+    def to_dict(self) -> Dict[str, Any]:
+        """Return a shallow copy of the raw contents (used for serialization).
+
+        Intentionally does not check the seal: the loader serializes a scope just
+        before sealing it, and callers should not reach for this directly.
+        """
+        return dict(self._data)
+
```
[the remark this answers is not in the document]
lilyydu: ^this would need to be manually invoked by the dev

packages/apps/src/microsoft_teams/apps/state/turn_state.py
```diff
@@ -0,0 +1,97 @@
...
+class TurnState(MutableMapping[str, Any]):
+    """One state scope for a single turn.
+
+    Behaves like a ``dict`` but adds two things the loader relies on:
+
+    * **Dirty tracking** — the loader only writes a scope back when it was
+      mutated, so an untouched scope costs nothing to "save".
+    * **Sealing** — at the end of a turn the scope is sealed; any later access
+      raises :class:`TurnStateSealedError`.
```
corinagum, line 26: Nested mutation doesn't mark the scope dirty, so ` _save_scope` skips the write:

```
turn 1 saved: {"oauth": {"github": {"pending": false}}}
turn 2: state.conversation["oauth"]["github"]["pending"] = True
        is_dirty: False  >  write dropped
turn 3: reads pending = False
```
Top-level assignment is fine, so the #554 example is safe, but PR 4's pending hints and PR 5's dedup marker are exactly this shape, and a dedup marker that doesn't persist means `sign_in` fires twice under concurrency. It also reads back correctly within the same turn, so it won't show up in manual testing.

C# doesn't have this issue, `TurnState.Get<T>()`  sets `IsDirty = true` on every return.

packages/apps/src/microsoft_teams/apps/state/container.py
```diff
@@ -0,0 +1,51 @@
...
+        """Clear both scopes and remove them from the backing store.
+
+        Clearing marks the scopes dirty+empty (so a later save deletes the keys), and the
+        injected deleter removes the keys immediately.
+        """
+        self.conversation.clear()
+        if self.user is not None:
+            self.user.clear()
+        if self._deleter is not None:
+            await self._deleter()
```
corinagum: C# deliberately throws here (`"State deletion is not available. Call UseState() during service registration."`). Maybe `RuntimeError` to match?

packages/apps/src/microsoft_teams/apps/state/turn_state.py
lilyydu: yep, this is the same comment as https://github.com/microsoft/teams.py/pull/553#discussion_r3760321428.
I'm evaluating how to handle this, we could save a footprint in the beginning, but trying to think of alternatives

packages/apps/src/microsoft_teams/apps/state/loader.py
```diff
@@ -0,0 +1,131 @@
...
+        save always targets the same keys the container was loaded from.
+        Empty-but-dirty scopes are deleted.
+        """
+        if not container.conversation_id:
+            raise ValueError("TurnStateContainer.conversation_id must be set to save state.")
+        if container.user is not None and not container.user_id:
+            raise ValueError("TurnStateContainer.user_id must be set to save user state.")
+        await self._save_scope(self.conversation_key(container.conversation_id), container.conversation)
+        if container.user is not None and container.user_id is not None:
+            await self._save_scope(self.user_key(container.conversation_id, container.user_id), container.user)
```
corinagum: Conversation is committed before the user scope is serialized, so if there's an error there's no rollback. #554 has a `finally`, which raises a `TypeError` but `RuntimeErrors` would get replaced. 

Serializing before write would make sure `TypeError`s fail ahead of time, don't need rollback, and `RuntimeError`s don't get replaced.

packages/apps/src/microsoft_teams/apps/state/loader.py
```diff
@@ -0,0 +1,131 @@
...
+        resolved = storage if storage is not None else self._options.storage
+        if resolved is None:
+            raise ValueError("TurnStateLoader requires a Storage backend (pass one explicitly or via StateOptions).")
+        self._storage: Storage[str, str] = resolved
+
+    @property
+    def options(self) -> StateOptions:
+        return self._options
+
+    def conversation_key(self, conversation_id: str) -> str:
```
corinagum, line 41: Segments should be escaped. id's on the service can carry `;`, `=`, etc. If this is a mirror to C# it probably also needs a fix. (same for `user_key`)

packages/apps/src/microsoft_teams/apps/state/loader.py
```diff
@@ -0,0 +1,131 @@
...
+        if not isinstance(parsed, dict):
+            return {}
+        blob = cast(Dict[str, Any], parsed)
+
+        if self._options.ttl is not None:
+            saved_at = blob.get("ts")
+            if not isinstance(saved_at, (int, float)):
+                return {}
+            if (time.time() - saved_at) > self._options.ttl:
+                return {}
```
corinagum: Recommend deleting the key when `_deserialize` detects expiry.

packages/apps/src/microsoft_teams/apps/state/options.py
```diff
@@ -0,0 +1,30 @@
...
+    has no native TTL concept.
+    """
+
+    storage: Optional[Storage[str, str]] = None
+    """Backing store for state blobs. When ``None`` the loader must be given one."""
+
+    key_prefix: str = "ts"
+    """Namespace prefix for scope keys (``{prefix}:conv:...`` / ``{prefix}:user:...``)."""
+
+    ttl: Optional[int] = None
```
corinagum: nit: This is more "sliding since last write", where C# allows picking between absolute vs. sliding. Should document as such.

packages/apps/src/microsoft_teams/apps/state/loader.py
lilyydu: I think `deserialize` should stay pure to parsing the blob. I updated `_load_scope` to delete expired and malformed stored blobs instead.

packages/apps/src/microsoft_teams/apps/state/turn_state.py
lilyydu: Updated dirty tracking to snapshot-compare the current scope contents against the loaded baseline, so nested dict/list mutations are now detected and persisted without dirtying read-only access. Agents SDK also does something similiar

Reason for not choosing to set isDirty per retrieval:
We expose TurnState as a dict, so the same read path is used for everything (reads and mutations). Two main issues - it over persists unnecessarily and it also refreshes the sliding TTL on read-only turns.

This is much cleaner and easier to do in C# - I'm gonna continue to see if there's a better design for this

packages/apps/src/microsoft_teams/apps/state/turn_state.py
```diff
@@ -0,0 +1,102 @@
...
+        return dict(self._data)
+
+    def _ensure_active(self) -> None:
+        if self._sealed:
+            raise TurnStateSealedError("TurnState has been sealed and can no longer be accessed.")
+
+    @staticmethod
+    def _fingerprint(data: Mapping[str, Any]) -> str:
+        canonical = json.dumps(data, sort_keys=True, default=repr, separators=(",", ":"))
+        return hashlib.sha256(canonical.encode("utf-8")).hexdigest()
```
corinagum, line 86: in #554, `_prepare_scope_save` runs inside `finally`, so a circular value would result in a turn exception. Maybe catch `(ValueError, TypeError)` in `_fingerprint` so `json.dumps` shows the error at save?

packages/apps/src/microsoft_teams/apps/state/options.py
```diff
@@ -0,0 +1,36 @@
...
```
corinagum: Should refresh `ts` on load-hit or document for PR 5.

packages/apps/src/microsoft_teams/apps/state/loader.py
```diff
@@ -0,0 +1,161 @@
...
+        return TurnState(cast(Dict[str, Any], data))
+
+    def _prepare_scope_save(
+        self,
+        key: str,
+        scope: TurnState,
+        pending_deletes: list[str],
+        pending_sets: list[tuple[str, str]],
+    ) -> None:
+        if not scope.is_dirty:
```
corinagum, line 132: `is_dirty` stays `True` after save, so a second save would rewrite everything. Recommend `save()` always returns to baseline.

packages/apps/src/microsoft_teams/apps/state/options.py
lilyydu: I prefer documenting. Refreshing on read would make read-only access extend TTL, which is intentionally avoided.

packages/apps/src/microsoft_teams/apps/state/turn_state.py
```diff
@@ -0,0 +1,102 @@
...
+import json
+from collections.abc import Iterator, Mapping, MutableMapping
+from typing import Any, Dict, Optional
+
+
+class TurnStateSealedError(RuntimeError):
+    """Raised when a sealed :class:`TurnState` is accessed after its turn ends."""
+
+
+class TurnState(MutableMapping[str, Any]):
```
MehakBindra, line 18: Can we add type derived keys like - https://github.com/microsoft/teams.net/blob/34f6f09ff89b3e280e47effb048f19cd4c22cef3/src/Microsoft.Teams.Apps/State/TurnState.cs#L154-L223 ?

lilyydu: mmm I am somewhat against this, type-derived keys make sense in C# cause it has strong generic APIs and typed object materialization. but we don't get the same value in PY unless we add a typed object mapping layer, I purposely avoided it by keeping state JSON-native

its more pythonic to do
`ctx.state.conversation["my_feature"] = {"count": 1}`

or
```python
MY_FEATURE_STATE = "my_feature"
ctx.state.conversation[MY_FEATURE_STATE] = {"count": 1}
```

packages/apps/src/microsoft_teams/apps/state/loader.py
lilyydu: updated so `save()` re-baselines dirty scopes after all pending storage operations succeed; failed serialization/storage keeps the scope dirty so the error isn’t hidden!

packages/apps/src/microsoft_teams/apps/state/turn_state.py
MehakBindra: Fair, makes sense

packages/apps/src/microsoft_teams/apps/state/options.py
corinagum: sgtm!

packages/apps/src/microsoft_teams/apps/state/loader.py
```diff
@@ -0,0 +1,169 @@
...
+        if user_id is not None:
+            await self._storage.async_delete(self.user_key(conversation_id, user_id))
+
+    async def _load_scope(self, key: str) -> TurnState:
+        raw = await self._storage.async_get(key)
+        if raw is None:
+            return TurnState()
+
+        blob = self._deserialize(raw)
+        if blob is None or self._is_expired(blob):
```
corinagum: Also deletes blobs that failed to parse. Maybe delete only on `_is_expired`

```
blob = self._deserialize(raw)
if blob is None:
    logger.debug("Discarding unreadable state blob at %s", key)
    return TurnState()
if self._is_expired(blob):
    await self._storage.async_delete(key)
    return TurnState()
```

lilyydu: I did this intentionally cause otherwise the corrupt entry will remain in storage and every turn, we'll repeatedly try to serialize a malformed blob

corinagum: This is more about the type narrowing, not malformed blobs. 
`async_get` can return a `dict`, which raises a type error, and `_deserialize` returns `None`, so it gets deleted. The store is valid, but because of type narrowing it's dropped. 

- A corrupt blob self-heals, and the empty scope is overwritten on the next save, so it doesn't accumulate.
- Deleting doesn't help the type case anyway: it fires every turn, data never accumulates, and it's destroyed each time.

Maybe accept `Storage[str, Any]` instead? Drop the cast in #554, so a `dict` will pass, and deleting malformed blobs is protected.

MehakBindra (approved):

MehakBindra (approved):

PR #554, opened 2026-08-10
feat: expose per-turn state (`ctx.state`, `App(state=...)`) and wire into dispatch

lilyydu:
## Summary

This is **PR 2 of 5** in the multi-connection OAuth stack.

PR #553 introduced the state foundation. This PR exposes state through `ctx.state`, adds the `App(state=...)` opt-in, and integrates state loading and persistence into activity dispatch.

State remains disabled by default. When disabled, `ctx.state` is `None` and no state storage operations occur.

### Public API

- Adds `ActivityContext.state`.
- Adds `state` to `AppOptions`.
- Re-exports:
  - `StateOptions`
  - `TurnState`
  - `TurnStateContainer`
  - `TurnStateSealedError`
  - `create_state_loader`

### Configuration

```python
# State disabled (default)
app = App()

# Use the app's shared storage
app = App(state=True)

# Use a dedicated provider or key prefix
app = App(
    state=StateOptions(
        storage=my_storage,
        key_prefix="my-app",
    )
)
```

`StateOptions` does not manage TTL. Expiration and other provider-specific behavior are configured directly on the selected storage provider.

When state resolves to in-memory `LocalStorage`, the SDK logs a warning because the data is process-local and lost on restart.

### Turn lifecycle

For each activity when state is enabled:

1. Conversation and user state are loaded before middleware and handlers run.
2. Handlers access them through:
   - `ctx.state.conversation`
   - `ctx.state.user`
3. Dirty state is persisted at the end of the turn, including when processing fails.
4. The container is always sealed after persistence, preventing post-turn access.
5. `on_activity_response` runs only after persistence succeeds.

Persistence failures are reported through `on_error`:

- If processing succeeded, the persistence failure propagates.
- If processing already failed or was cancelled, that original error remains authoritative.
- Failures from error-reporting plugins cannot replace the original processing error.

Existing `StreamCancelledError` behavior remains: the stream is closed and the activity returns a 200 response.

### Storage resolution

`create_state_loader()` centralizes resolution of:

- `None` / `False` → state disabled
- `True` → app storage with default state options
- `StateOptions(...)` → configured prefix and optional dedicated storage

### Example

Adds `examples/state/`, demonstrating:

- A conversation-scoped message counter
- User-scoped profile data
- Automatic loading and persistence
- `LocalStorage` development behavior

### DIFF
--- a/examples/state/src/main.py
+++ b/examples/state/src/main.py
@@ -34,7 +34,7 @@ async def handle_message(ctx: ActivityContext[MessageActivity]) -> None:
     count = ctx.state.conversation.get("message_count", 0) + 1
     ctx.state.conversation["message_count"] = count
 
-    # User scope: per-sender. `user` is None only when the activity has no sender.
+    # User scope: per-sender. `user` is None when the activity has no usable, non-empty sender ID.
     greeting = ""
     if ctx.state.user is not None:
         if "name" not in ctx.state.user:
--- a/examples/state/src/main.py
+++ b/examples/state/src/main.py
@@ -1,0 +1,50 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+import asyncio
+import logging
+
+from microsoft_teams.api import MessageActivity
+from microsoft_teams.apps import ActivityContext, App
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+# `state=True` enables the per-turn state layer using the app's storage.
+#
+# With no other storage configured the app falls back to in-memory
+# `LocalStorage`, so state is scoped to a single process and lost on restart —
+# the SDK logs a warning to that effect. For production, pass a durable store
+# via `App(state=StateOptions(storage=...))`
+#
+# State is loaded before each turn and saved automatically after it, then
+# sealed — reading or writing `ctx.state` after the handler returns raises
+# `TurnStateSealedError`.
+app = App(state=True)
+
+
+@app.on_message
+async def handle_message(ctx: ActivityContext[MessageActivity]) -> None:
+    """Track a per-conversation message count and a per-user first-seen name."""
+    assert ctx.state is not None
+
+    # Conversation scope: shared by everyone in the chat/channel.
+    count = ctx.state.conversation.get("message_count", 0) + 1
+    ctx.state.conversation["message_count"] = count
+
+    # User scope: per-sender. `user` is None only when the activity has no sender.
+    greeting = ""
+    if ctx.state.user is not None:
+        if "name" not in ctx.state.user:
+            ctx.state.user["name"] = ctx.activity.from_.name
+            greeting = f"Nice to meet you, {ctx.activity.from_.name}! "
+        else:
+            greeting = f"Welcome back, {ctx.state.user['name']}! "
+
+    await ctx.send(f"{greeting}This conversation has seen {count} message(s).")
+
+
+if __name__ == "__main__":
+    asyncio.run(app.start())
--- a/packages/apps/src/microsoft_teams/apps/app_process.py
+++ b/packages/apps/src/microsoft_teams/apps/app_process.py
@@ -4,7 +4,6 @@ Licensed under the MIT License.
 """
 
 import logging
-import sys
 from time import perf_counter
 from typing import TYPE_CHECKING, Any, Awaitable, Callable, Dict, List, Optional, TypeGuard, Union, cast
 
@@ -291,6 +290,7 @@ class ActivityProcessor:
         if not self.event_manager:
             raise ValueError("EventManager was not initialized properly")
 
+        processing_completed = False
         try:
             await self._load_turn_state(activityCtx, activity)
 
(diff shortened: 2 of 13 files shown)


packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -297,7 +303,26 @@ async def route(ctx: ActivityContext[ActivityBase]) -> Optional[Any]:
...
+        except Exception as error:
+            await self.event_manager.on_error(ErrorEvent(error=error, activity=activity), plugins)
+            raise
+        finally:
+            error_in_flight = sys.exc_info()[1] is not None
+            try:
+                await self._persist_turn_state(activityCtx)
+            except Exception as error:
+                try:
+                    await self.event_manager.on_error(ErrorEvent(error=error, activity=activity), plugins)
```
MehakBindra, line 320: Shouldn't the error passed to the plugin system and the error raised be same ?

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -297,7 +303,26 @@ async def route(ctx: ActivityContext[ActivityBase]) -> Optional[Any]:
...
             else:
                 response = InvokeResponse[Any](status=200, body=middleware_result)
+        except StreamCancelledError:
+            logger.debug("Activity processing was cancelled (stream stopped)")
+            await activityCtx.stream.close()
+            response = InvokeResponse[Any](status=200)
+        except Exception as error:
+            await self.event_manager.on_error(ErrorEvent(error=error, activity=activity), plugins)
+            raise
+        finally:
```
corinagum, line 315: This will silently swallow save failures. If `process_activity` is awaited inside an unrelated `except ValueError:`, `InvokeResponse(status=200)` is returned.

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -297,7 +303,26 @@ async def route(ctx: ActivityContext[ActivityBase]) -> Optional[Any]:
                 response = cast(InvokeResponse[Any], middleware_result)
             else:
                 response = InvokeResponse[Any](status=200, body=middleware_result)
+        except StreamCancelledError:
```
corinagum, line 307: I think we need a cancellation + persistence test. Which one wins? 

If `execute_middleware_chain` raises a `StreamCancelledError` with `save`, there will be a `RuntimeError("save failed")` instead of 200.

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -312,12 +337,38 @@ async def route(ctx: ActivityContext[ActivityBase]) -> Optional[Any]:
...
+    async def _load_turn_state(self, ctx: ActivityContext[ActivityBase], activity: ValidatedActivity) -> None:
+        """Load per-turn state onto ``ctx.state`` when state is enabled.
+
+        Loads both the conversation scope and the user scope (keyed by the
+        activity's ``from`` identity). A no-op when state is disabled, leaving
+        ``ctx.state`` as ``None``.
+        """
+        if self.state_loader is None:
+            return
+        ctx.state = await self.state_loader.load(activity.conversation.id, activity.from_.id)
```
corinagum: Maybe add `or None` for `activity.from_.id`? `Account.id` isn't constrained to non-empty, so `from.id == ""` produces user scope with `user_id=""`. `TurnStateLoader.save` will produce `TurnStateContainer.user_id must be set to save user state`.

packages/apps/tests/test_app_process.py
```diff
@@ -846,3 +889,246 @@ async def test_process_activity_raises_exception(self, activity_processor):
...
+            await activity_processor.process_activity([], self._message_event())
+
+        assert exc_info.value is handler_error
+        assert container.conversation.is_sealed is True
+        errors = [call.args[0].error for call in activity_processor.event_manager.on_error.await_args_list]
+        assert errors == [handler_error, save_error]
+        activity_processor.event_manager.on_activity_response.assert_not_awaited()
+
+    @pytest.mark.asyncio
+    async def test_state_save_failure_does_not_mask_task_cancellation(self, activity_processor):
```
corinagum: Maybe name to `test_state_save_failure_does_not_mask_base_exception` - and add a case that cancels the task when save is in-progress.

examples/state/src/main.py
```diff
@@ -0,0 +1,50 @@
...
+@app.on_message
+async def handle_message(ctx: ActivityContext[MessageActivity]) -> None:
+    """Track a per-conversation message count and a per-user first-seen name."""
+    assert ctx.state is not None
+
+    # Conversation scope: shared by everyone in the chat/channel.
+    count = ctx.state.conversation.get("message_count", 0) + 1
+    ctx.state.conversation["message_count"] = count
+
+    # User scope: per-sender. `user` is None only when the activity has no sender.
```
corinagum: This code never runs: right now `ctx.state.user` is never `None`.

packages/apps/src/microsoft_teams/apps/app_process.py
lilyydu: They’re the same when only saving fails. If the handler already failed, we keep raising that error and report the save error separately so the original error isn’t lost

examples/state/src/main.py
lilyydu: Good catch. I changed state loading to treat an empty sender ID as no user ID (`activity.from_.id or None`). That makes `ctx.state.user` `None`, so this guard is now meaningful

corinagum (approved):
<3

PR #556, opened 2026-08-11
Consolidate message interaction examples

corinagum:
## Summary

- replace the separate quoting, threading, and reactions examples with `interacting-with-messages`
- keep each concept and command in focused Python modules with explicit handler composition
- demonstrate reactive, proactive, batch, and manual message interaction flows
- update workspace metadata plus consistent help and README command descriptions

## Validation

- `uv run ruff format --check examples/interacting-with-messages`
- `uv run ruff check examples/interacting-with-messages`
- `uv run pyright examples/interacting-with-messages`
- manually tested help rendering, quote reply/message, reaction add/remove/proactive, and reactive/proactive/manual threading in Teams


examples/interacting-with-messages/src/threading_handlers.py
```diff
@@ -0,0 +1,48 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from microsoft_teams.api import MessageActivity
+from microsoft_teams.apps import ActivityContext, App, to_threaded_conversation_id
+
+
```
lilyydu, line 9: nit: should we name this `threading.py` to match the others?

examples/interacting-with-messages/pyproject.toml
```diff
@@ -1,7 +1,7 @@
 [project]
-name = "reactions"
+name = "interacting-with-messages"
```
lilyydu, line 2: nit: I'm blanking on ideas for this but I think it might be good to specify more in the name, otherwise it sounds simple (vs what it showcases is pretty important!)

examples/interacting-with-messages/src/threading_handlers.py
corinagum: It looks like it may be better to leave this as `threading_handlers`. In py, `threading.py` would become a top-level module named `threading`. Since `src/` is first on `sys.path` when running `src/main.py`, it can override the standard-library `threading` module. Leaving as-is.

examples/interacting-with-messages/pyproject.toml
corinagum: Yeah I considered the same thing. I think it's fine - the README clearly states in the first sentence what is being showcased, and we discussed this revamp with the PMs who all seemed fine with the proposed name. We also don't want to do something like `quoting-threading-reacting` since we want to keep this open to similar features being added on in the future.

lilyydu (approved):

PR #561, opened 2026-08-13
feat: add multi-connection OAuth registry and flow APIs

lilyydu:
## Summary

This PR adds the public registry and flow APIs for using multiple named OAuth connections in `teams.py`.

It is stacked on #554 and is intentionally limited to the **registration and connection-bound API layer**. Callback routing, pending sign-in attribution, and state-backed deduplication are deferred to follow-up PRs.

The API follows the C# SDK's `OAuthFlow` / `AddOAuthFlow(...)` model while remaining additive and backward-compatible with the existing app-level default OAuth connection.

### Register named OAuth flows

Added `OAuthFlow` and a case-insensitive, insertion-ordered `OAuthFlowRegistry`.

```python
github = app.add_oauth_flow(
    "github",
    oauth_card_text="Sign in with GitHub",
    sign_in_button_text="Continue",
)

graph = app.add_oauth_flow("graph")

same_github_flow = app.get_oauth_flow("GITHUB")
```

- `App.add_oauth_flow(...)` registers and returns an `OAuthFlow`.
- `App.get_oauth_flow(...)` retrieves a registered flow case-insensitively.
- Duplicate connection names raise `ValueError` case-insensitively.
- Missing connections raise `ValueError` with the registered connection names.
- `OAuthFlow` and `OAuthFlowRegistry` are exported from `microsoft_teams.apps`.

### Add connection-bound flow operations

Each `OAuthFlow` delegates OAuth operations to its own connection:

```python
await github.sign_in(ctx)
await github.sign_out(ctx)

token = await github.get_token(ctx)
signed_in = await github.is_signed_in(ctx)
```

`OAuthFlow.sign_in(...)` preserves caller-provided card text and button text but always forces the flow's connection name, preventing options for one connection from being used through another flow.

Flows also expose `on_signin(...)` and `on_signin_failure(...)` handler registration. Invocation of those per-flow handlers will be wired by the callback-routing layer in a follow-up PR.

### Add per-connection `ActivityContext` helpers

- `ctx.sign_out(connection_name=...)` signs out a specific connection.
- `ctx.get_user_token(connection_name=...)` retrieves the token for a specific connection.
- Omitting `connection_name` preserves the existing default-connection behavior.
- `get_user_token(...)` returns `None` only when Token Service returns `404`.
- Non-404 Token Service failures propagate instead of being reported as a signed-out user.
- `ctx.get_token_status()` makes one request and returns status for all bot OAuth connections.

### Add a structured sign-in failure event

Added `SignInFailureEvent`, containing:

- the activity context;
- connection name;
- failure code;
- failure message.

The existing OAuth failure handler emits both the existing `ErrorEvent` for backward compatibility and the new `sign_in_failure` event for structured handling.

## Backward compatibility

Existing single-connection OAuth APIs continue to use the app-level default connection when no connection name is provided. Registering flows is optional, so existing applications do not need to change.

## Deferred to follow-up PRs

This PR does **not** yet:

- route token exchange, verify-state, or failure callbacks to a registered flow;
- invoke per-flow success or failure handlers;
- persist pending sign-in attribution in turn state;
- add state-backed callback deduplication;
- update the OAuth example to demonstrate end-to-end multi-connection behavior.

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/events/types.py
+++ b/packages/apps/src/microsoft_teams/apps/events/types.py
@@ -11,6 +11,7 @@ from microsoft_teams.api import (
     ConversationReference,
     InvokeResponse,
     SentActivity,
+    SignInFailureInvokeActivity,
     SignInTokenExchangeInvokeActivity,
     SignInVerifyStateInvokeActivity,
     TokenProtocol,
@@ -117,3 +118,13 @@ class SignInEvent:
         ActivityContext[SignInTokenExchangeInvokeActivity],
     ]
     token_response: TokenResponse
+
+
+@dataclass
+class SignInFailureEvent:
+    """Event emitted when a sign-in (silent SSO) attempt fails."""
+
+    activity_ctx: ActivityContext[SignInFailureInvokeActivity]
+    connection_name: Optional[str] = None
+    code: Optional[str] = None
+    message: Optional[str] = None
--- a/packages/apps/src/microsoft_teams/apps/events/types.py
+++ b/packages/apps/src/microsoft_teams/apps/events/types.py
@@ -118,6 +118,7 @@ class SignInEvent:
         ActivityContext[SignInTokenExchangeInvokeActivity],
     ]
     token_response: TokenResponse
+    connection_name: Optional[str] = None
 
 
 @dataclass
--- a/packages/apps/src/microsoft_teams/apps/oauth_flow.py
+++ b/packages/apps/src/microsoft_teams/apps/oauth_flow.py
@@ -1,0 +1,115 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+import logging
+from collections import OrderedDict
+from dataclasses import replace
+from typing import Any, Awaitable, Callable, Iterator, List, Mapping, Optional
+
+from .events import SignInEvent, SignInFailureEvent
+from .routing import ActivityContext, SignInOptions
+
+logger = logging.getLogger(__name__)
+
+SignInHandler = Callable[[SignInEvent], Awaitable[None]]
+SignInFailureHandler = Callable[[SignInFailureEvent], Awaitable[None]]
+
+DEFAULT_OAUTH_CARD_TEXT = SignInOptions().oauth_card_text
+DEFAULT_SIGN_IN_BUTTON_TEXT = SignInOptions().sign_in_button_text
+
+
+class OAuthFlow:
+    """One named OAuth connection, plus the handlers attached to it.
+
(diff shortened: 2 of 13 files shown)


packages/apps/tests/test_oauth_flow.py
```diff
@@ -0,0 +1,219 @@
...
+    def test_on_signin_registers_and_returns_handler(self) -> None:
+        flow = OAuthFlow("graph")
+
+        async def handler(event) -> None:
+            pass
+
+        returned = flow.on_signin(handler)
+
+        assert returned is handler
+        assert flow._on_signin == [handler]
```
corinagum: This is private, right? (same is done below). I think these lines should be dropped and add a test that registers the handler, signs in, and check that it fired.

packages/apps/src/microsoft_teams/apps/events/types.py
```diff
@@ -117,3 +118,13 @@ class SignInEvent:
         ActivityContext[SignInTokenExchangeInvokeActivity],
     ]
     token_response: TokenResponse
+
+
+@dataclass
+class SignInFailureEvent:
```
corinagum, line 125: Is `SignInEvent` with `connection_name` for a later PR? TS 744 adds it in that PR.

packages/apps/src/microsoft_teams/apps/oauth_flow.py
```diff
@@ -0,0 +1,115 @@
...
+    truthiness all work without extra code.
+    """
+
+    def __init__(self) -> None:
+        self._flows: "OrderedDict[str, OAuthFlow]" = OrderedDict()
+
+    def __getitem__(self, connection_name: str) -> OAuthFlow:
+        return self._flows[connection_name.lower()]
+
+    def __iter__(self) -> Iterator[str]:
```
corinagum, line 100: Hmm `self._flows` was lowercased, so a flow registered as `GitHub` will come back as `github`. Using `flow.connection_name` will keep the casing.

corinagum (approved):
`app_oauth.py:64-69` says "Token verification will likely fail" With registered non-default connections now being supported, SSO will send a false alarm. Maybe update to check whether a flow is registered for that connection?

packages/apps/tests/test_oauth_flow.py
lilyydu: I'll remove them for now cause the callbacks are wired in the next PR!

packages/apps/src/microsoft_teams/apps/events/types.py
lilyydu: Ty! I forgot to push this change cause the routing work is for PR 4

corinagum (approved):

PR #573, opened 2026-08-25
feat(oauth): route sign-in callbacks on turn state

lilyydu:
**Stack:** #553 (state foundation) → #554 (public state API / dispatch lifecycle) →
#561 (OAuthFlow + registry) → **this PR (4/5)** → PR 5 (token-exchange dedup)

This PR makes callback routing connection-aware.

## Problem

The three sign-in callbacks carry very different amounts of information:

| Callback | Carries connection name? |
| --- | --- |
| `signin/tokenExchange` | ✅ yes, in `activity.value.connection_name` |
| `signin/verifyState` | ❌ no — only an opaque magic code |
| `signin/failure` | ❌ no — only a failure code/message |

Token exchange can be routed directly. The other two cannot, so the app has to
remember which connection it last offered a sign-in on.

## Approach

**Explicit routing where possible.** `signin/tokenExchange` routes purely on the
connection name in the payload — no state required, and it works with state
disabled.

**Pending markers for connection-less callbacks.** When a sign-in card is sent,
`ctx.sign_in()` records the attempt in per-turn *user* state under private
reserved keys — up to two per connection, each holding an ISO 8601 UTC timestamp:

```python
"__oauth:pending:GitHub":     "2026-08-26T20:48:20.214000+00:00"
"__oauth:pending:sso:GitHub": "2026-08-26T20:48:20.214000+00:00"  # only when SSO was offered
```

**Probing with fallback.** `signin/verifyState` tries attributed flows first, then
the remaining registered flows, then the legacy default connection, stopping at
the first connection that returns a token. Attribution makes the normal path a
single Token Service call; the fan-out is the fallback for missing or stale
markers. It is intentionally uncapped — dropping candidates would turn a slow
sign-in into a silently failed one.

**Failure attribution.** `signin/failure` is attributed to the newest attempt that
actually offered silent SSO. With no attribution it falls back to notifying every
registered flow.

**SSO markers are retired, not deleted.** After a silent-SSO failure Teams still
renders the sign-in button on the same card, so the sign-in is still pending even
though its SSO attempt is spent. Only the `sso:` key is dropped: the attempt can
no longer re-attribute a second failure, but a follow-up `verifyState` still
routes straight to the right connection, and it expires on its original schedule.

## Behavior

- `SignInEvent.connection_name` and `SignInFailureEvent.connection_name` are now
  populated, preserving the registry's original casing (lookup is
  case-insensitive; `"github"` resolves the flow registered as `"GitHub"`).
- Global `sign_in` / `sign_in_failure` events fire before per-flow handlers, and
  are now **awaited**, so an async global handler is guaranteed to complete first.
- Legacy single-connection apps are unchanged: with no flows registered, routing
  falls through to `default_connection_name` exactly as before.
- Legacy and registered modes are not mutually exclusive — a registered default
  flow and unregistered connections coexist.
- Error precedence and Token Service status handling are untouched: only 404
  continues probing, 400/412 return 412, and any other status propagates
  unchanged and still emits `error`. No service failure is masked as a
  logged-out/unknown callback.
- Any stored value that is not a parseable ISO 8601 timestamp, along with stale,
  future-dated and unknown-connection markers, is discarded and routing falls
  back to the default connection.

## State is optional

State is only required for the two callbacks that genuinely cannot self-identify.
With state disabled, `tokenExchange` still routes correctly and `verifyState` /
`failure` use the registered-flow fallback. No generic TTL knobs are introduced —
storage expiry remains provider-configured; the 5-minute freshness window is
OAuth-specific and internal.

## Race fix: markers are flushed before the card is sent

Turn state is normally persisted at end-of-turn, but a token-exchange callback can
arrive *before* the turn that sent the card has finished. `ctx.sign_in()` now
flushes the marker to storage before sending the card, and rolls it back (best
effort, without masking the caller's error) if the send fails.

## Container identity is bound at load

`TurnStateContainer.conversation_id` and `.user_id` describe *which* state was
loaded. Reassigning them after construction cannot move the data and would make
the container save under a key that no longer matches its contents, so they are
now read-only: private backing fields behind getter-only properties, rejected at
both runtime and type-check time.

The contained `TurnState` objects stay fully mutable and `seal()`, `delete()` and
`_save()` are unchanged. The constructor signature is identical (`conversation_id=`
/ `user_id=` keywords), so no call site moved. `TurnStateContainer` is no longer a
`@dataclass` — `__eq__` and `__repr__` are hand-written to reproduce the generated
ones exactly, including excluding the injected hooks.

## Supporting changes

- `EventEmitter.emit_async()` — awaits all handlers. Purely additive; existing
  `emit()` stays fire-and-forget and `EventEmitterProtocol` is unchanged, so
  external implementers are unaffected.
- The `emit()` refactor fixes two latent bugs: awaitables returned by callables
  that aren't `async def` (e.g. `functools.partial`, `__call__`) were created and
  never awaited, and an async `once` handler could fire twice.
- `TurnStateContainer._save()` / `TurnState._mark_dirty()` — internal hooks for
  the mid-turn flush and its rollback.

## Not in this PR

Token-exchange deduplication (in-flight guard + completed marker) is PR 5.
Duplicate `tokenExchange` callbacks still run the exchange twice here, and there
are tests pinning that so the change in PR 5 is visible.

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/app.py
+++ b/packages/apps/src/microsoft_teams/apps/app.py
@@ -459,6 +459,15 @@ class App(ActivityHandlerMixin):
         Returns:
             The registered ``OAuthFlow``.
 
+        Notes:
+            Enable per-turn state with ``App(state=True)`` (or ``StateOptions``)
+            to record pending connection hints. Connection-less verify-state
+            callbacks probe hinted flows first, then the remaining registered
+            flows and legacy default connection. Sign-in failures use pending
+            silent-SSO hints for attribution and notify registered flows as a
+            fallback. Token-exchange callbacks carry their connection name and
+            route correctly without state.
+
         Raises:
             ValueError: if a flow for this connection is already registered
                 (connection names are case-insensitive).
--- a/packages/apps/src/microsoft_teams/apps/app.py
+++ b/packages/apps/src/microsoft_teams/apps/app.py
@@ -149,6 +149,7 @@ class App(ActivityHandlerMixin):
             fetch_user_token=self.options.fetch_user_token,
             agent365_baggage_options=self.options.telemetry.get("agent365") if self.options.telemetry else None,
             state_loader=self._state_loader,
+            oauth_registry=self._oauth_registry,
         )
         self.event_manager = EventManager(self._events)
         self.activity_processor.event_manager = self.event_manager
--- a/packages/apps/src/microsoft_teams/apps/app.py
+++ b/packages/apps/src/microsoft_teams/apps/app.py
@@ -461,24 +461,56 @@ class App(ActivityHandlerMixin):
             The registered ``OAuthFlow``.
 
         Notes:
-            Enable per-turn state with ``App(state=True)`` (or ``StateOptions``)
-            to record pending connection hints. Connection-less verify-state
-            callbacks probe hinted flows first, then the remaining registered
-            flows and legacy default connection. Sign-in failures use pending
-            silent-SSO hints for attribution and notify registered flows as a
-            fallback. Token-exchange callbacks carry their connection name and
-            route correctly without state.
+            Registering a flow turns on per-turn state automatically when the
+            ``state`` option was omitted, because connection-less callbacks need
+            somewhere to record which connection a sign-in started on. Pass
+            ``state=True`` or a ``StateOptions`` to choose the storage yourself,
+            or ``state=False`` to opt out — an explicit choice is never
+            overridden. With state off, pending hints fall back to a bounded
+            process-local cache, so a callback handled by another instance
+            cannot be attributed.
+
+            Connection-less verify-state callbacks probe hinted flows first,
+            then the remaining registered flows and legacy default connection.
+            Sign-in failures use pending silent-SSO hints for attribution and
+            notify registered flows as a fallback. Token-exchange callbacks
+            carry their connection name and route correctly without state.
 
         Raises:
             ValueError: if a flow for this connection is already registered
                 (connection names are case-insensitive).
         """
-        return self._oauth_registry.add(
+        flow = self._oauth_registry.add(
             OAuthFlow(
                 connection_name,
                 oauth_card_text=oauth_card_text,
                 sign_in_button_text=sign_in_button_text,
             )
+        )
+        # Only after the registry accepts it: a rejected duplicate must not be
+        # able to switch state on as a side effect.
+        self._enable_state_for_oauth_if_unset()
+        return flow
+
+    def _enable_state_for_oauth_if_unset(self) -> None:
+        """Turn on per-turn state for OAuth when the app did not say either way.
+
+        A registered flow needs durable pending attribution to route
+        connection-less ``signin/verifyState`` and ``signin/failure`` callbacks,
+        so state defaults on once a flow exists — matching the C# and TypeScript
+        SDKs. ``options.state`` still records what the caller passed: ``None``
+        means "unset", and anything else, including ``False``, is an explicit
+        decision this leaves alone. An already-resolved loader is also left
+        alone, so registering a second flow does not rebuild it.
+        """
+        if self.options.state is not None or self._state_loader is not None:
+            return
+
+        self._state_loader = create_state_loader(True, self.storage)
+        self.activity_processor.state_loader = self._state_loader
+        logger.debug(
+            "Enabled per-turn state because an OAuth flow was registered and no state option was set. "
+            "Pass state=False to opt out, or state=True/StateOptions to choose the storage."
         )
 
     def get_oauth_flow(self, connection_name: str) -> OAuthFlow:
--- a/packages/apps/src/microsoft_teams/apps/app_oauth.py
+++ b/packages/apps/src/microsoft_teams/apps/app_oauth.py
@@ -186,6 +186,26 @@ class OauthHandlers:
         logs the failure details and emits an error event so developers are
         notified rather than having the failure silently swallowed.
 
+        **Connection attribution.** A ``signin/failure`` callback does not carry a
+        connection name, so the failed connection is recovered from the pending
+        sign-in recorded when the flow started — durable state when state is
+        enabled, otherwise a short-lived process-local cache. Three outcomes:
+
+            - No flows registered (legacy mode): the default connection is the only
+              connection there is, so it is reported as the failed one.
+            - A pending sign-in resolves: that flow is reported, and only its
+              failure handlers run.
+            - Flows are registered but nothing resolves — most commonly because the
+              callback reached a different process than the one that started the
+              sign-in, and state is not enabled to bridge them. The failed
+              connection is genuinely unknown, so the global
+              ``SignInFailureEvent.connection_name`` is ``None`` and telemetry omits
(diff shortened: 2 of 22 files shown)


packages/apps/src/microsoft_teams/apps/oauth_flow.py
```diff
@@ -57,6 +62,14 @@ def on_signin_failure(self, func: SignInFailureHandler) -> SignInFailureHandler:
         self._on_signin_failure.append(func)
         return func
 
+    async def _invoke_signin_handlers(self, event: SignInEvent) -> None:
+        for handler in tuple(self._on_signin):
+            await handler(event)
+
+    async def _invoke_signin_failure_handlers(self, event: SignInFailureEvent) -> None:
+        for handler in tuple(self._on_signin_failure):
+            await handler(event)
```
corinagum, line 71: Maybe match what `emit_async` does? Await only if the result is await-able. Otherwise they await unconditionally so a good handler could be blocked by a bad one. Globals will isolate them. Routing through an emitter would remove duplication.

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -177,7 +197,11 @@ async def sign_in_failure(
         """
         activity = ctx.activity
         next_handler = ctx.next
-        connection_name = self.default_connection_name
+        pending_flows = self.oauth_registry._pending_flows(  # pyright: ignore[reportPrivateUsage]
+            ctx, sso_only=True
+        )
+        target_flow = pending_flows[0] if pending_flows else None
+        connection_name = target_flow.connection_name if target_flow is not None else self.default_connection_name
```
corinagum, line 204: This would name the default connection even if it's not a part of the failure. Maybe pass `None` as the event's connection name when there's no hint, and keeping `default_connection_name` for the span attribute. `SignInFailureEvent.connection_name` is already `Optional[str]`, so `None` would mean 'couldn't attribute' instead of naming a bystander.

Separately, with state disabled and three flows registered, a single SSO failure will fire three failure callbacks. Maybe add to explicit documentation?

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -263,35 +333,38 @@ async def sign_in_verify_state(
...
+                            )
+                            continue
+                        logger.error(
+                            f"Error verifying sign-in state for user {activity.from_.id} in conversation"
+                            f"{activity.conversation.id}: {e}"
+                        )
+                        if status not in (400, 412):
+                            await self.event_emitter.emit_async(
+                                "error",
+                                ErrorEvent(error=e, context={"activity": activity}),
```
corinagum, line 366: Hmm worth coordinating with Mehak for TS. PY and TS will resolve multi-connection magic-code sign-in differently.

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -236,7 +277,36 @@ async def sign_in_verify_state(
...
+                continue
+            seen_connections.add(key)
+            candidates.append((flow.connection_name, flow))
+
+        default_flow = self.oauth_registry.get(self.default_connection_name)
+        default_connection_name = (
+            default_flow.connection_name if default_flow is not None else self.default_connection_name
+        )
+        if default_connection_name.lower() not in seen_connections:
+            candidates.append((default_connection_name, default_flow))
```
MehakBindra, line 334: Should registered OAuth flows be mutually exclusive with the legacy default, as in TypeScript? it's simlpler behavior to track, one or the other

MehakBindra: also creates a clean migration boundary for eventually removing the deprecated legacy path

packages/apps/src/microsoft_teams/apps/oauth_state.py
```diff
@@ -0,0 +1,250 @@
...
+    sso_offered: bool
+
+
+def record_pending_oauth_sign_in(
+    state: Optional[TurnStateContainer],
+    connection_name: str,
+    *,
+    sso_offered: bool,
+) -> None:
+    if state is None or state.user is None:
```
MehakBindra, line 51: - **TypeScript and .NET:** use persisted state when available, otherwise a **per-process in-memory pending-flow map**.
- **Python:** has no in-memory fallback for pending attribution; without state it immediately probes every flow.
worth aligning ?

packages/apps/src/microsoft_teams/apps/routing/activity_context.py
MehakBindra, line 84: we should deprecate this ?

MehakBindra: While we still have override_sign_in_activity, we should use it

MehakBindra: only 404 should propagate to oauth card, rest throw ?

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -301,17 +374,54 @@ async def sign_in_verify_state(
                             span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_result, result)
                             return InvokeResponse(status=status)
                         result = APP_OAUTH_RESULTS.failure
-                    else:
+                        span.set_attribute(APP_ATTRIBUTE_NAMES.invoke_response_status, 412)
+                        span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_result, result)
+                        return InvokeResponse(status=412)
+                    except Exception as e:
```
MehakBindra, line 419: for non http errors should we throw and not send 412 ?

packages/apps/src/microsoft_teams/apps/app.py
```diff
@@ -459,6 +459,15 @@ def add_oauth_flow(
         Returns:
             The registered ``OAuthFlow``.
 
+        Notes:
+            Enable per-turn state with ``App(state=True)`` (or ``StateOptions``)
```
MehakBindra: should we enable state by default for oauth to align with TS and .NET, it's a pretty major difference

packages/apps/src/microsoft_teams/apps/app_oauth.py
lilyydu: I kept these as co-existing so we're still backwards compatible. It shouldn't be too complicated to deprecate the legacy later on

packages/apps/src/microsoft_teams/apps/app.py
lilyydu: .NET doesn't automatically enable state - only if it's explicit through `UseState()`, otherwise it uses private memory. I think this makes more sense to me, what are your thoughts?

^ rediscussed, C# does autoenable, will update!

packages/apps/src/microsoft_teams/apps/oauth_flow.py
lilyydu: updating so that we only await for awaitable results and isolate bad handlers! didn't go through `emity_async` because it would make async handlers concurrent and change the ordering semantics

corinagum (approved):

PR #576, opened 2026-08-26
feat(oauth): dedup, update oauth/graph samples, add dep markers on legacy methods

lilyydu:
PR 5 of the multi-connection OAuth stack.

## Problem

Teams fans `signin/tokenExchange` out to every endpoint the user is signed in on, so one login reaches the bot several times — each copy redeems the token, emits `sign_in`, and runs the flow's success handlers.

## Dedup

Two layers, mirroring the C# SDK:

1. **In-flight guard** — keyed by exchange id. A duplicate arriving mid-flight awaits the original and returns an equivalent result, including the `412` body on failure. Released on settle, so retrying a *failed* exchange still works.
2. **Completed marker** — after success the id is recorded so a *late* duplicate short-circuits to `200`. Held in memory and in conversation state (`__oauth:exchange:{id}`) for cross-instance coverage, and **never cleared** — otherwise a late duplicate runs as a fresh exchange.

`exchange_token`, the `sign_in` emission, and the success handlers each run exactly once. In-flight is checked first, since the marker is stamped at redemption while the owner still has callbacks to run. Check-and-insert has no `await` between lookup and insert — that is what makes it atomic on one event loop. Both in-memory structures are bounded (5-minute TTL, 1000-entry cap, pruned on access).

Deduplicated requests return before `next()`, so concurrent duplicates call it once in total, matching TS. `signin/verifyState` and `signin/failure` are deliberately **not** deduplicated — the verify code is single-use, failure is one informational notice. Tests pin both.

## Also included

- **`ctx.get_token_status` → `ctx.get_connection_status`** — it shadowed `api.users.get_token_status` while behaving differently (the context method adds a silent-SSO correction pass). Matches `getConnectionStatus` in teams.ts. Added in #561, after every published tag, so no release exposed the old name.
- **`OAuthFlow.sign_out` / `get_token` / `is_signed_in` now call `api.users.*` directly** instead of the deprecated context methods, as teams.ts does. Verified equivalent: same request params, logs, return values, and 404 handling. `sign_in` still delegates — extracting it is a follow-up.
- **Legacy single-connection surface deprecated in docstrings only**, no decorators, so the supported flow API never warns callers about an API they did not touch.
- **`examples/oauth`** now shows multi-connection; **`examples/graph`** moved to the flow API, fixing its own copy of the bug — it built a Graph client from `ctx.user_token`, i.e. whichever connection last exchanged.

NOTE: Once we fully remove the old logic, we need to extract`sign_in` into a shared module

### DIFF
--- a/packages/apps/src/microsoft_teams/apps/app_oauth.py
+++ b/packages/apps/src/microsoft_teams/apps/app_oauth.py
@@ -144,6 +144,16 @@ class OauthHandlers:
         api = ctx.api
         next_handler = ctx.next
         connection_name = activity.value.connection_name
+        # Resolved before the ``try`` so the ``finally`` below can always report it. The
+        # lookup is a plain dict access; computing it inside the ``try`` left it unbound
+        # whenever anything above it raised, which would turn the metric write in
+        # ``finally`` into an ``UnboundLocalError`` masking the original exception.
+        flow = self.oauth_registry.get(connection_name)
+        # Teams echoes back whatever casing the sign-in card carried, so telemetry keyed
+        # on the raw name splits one connection into several series. ``connection_name``
+        # stays raw: it is what goes on the wire to the Token Service, and rewriting that
+        # would be a behavior change rather than a telemetry fix.
+        event_connection_name = flow.connection_name if flow is not None else connection_name
         result = APP_OAUTH_RESULTS.failure
         started_at = perf_counter()
         try:
@@ -152,11 +162,8 @@ class OauthHandlers:
                 record_exception=False,
                 set_status_on_exception=False,
             ) as span:
-                span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_connection, connection_name)
+                span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_connection, event_connection_name)
                 span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_operation, APP_OAUTH_OPERATIONS.token_exchange)
-
-                flow = self.oauth_registry.get(connection_name)
-                event_connection_name = flow.connection_name if flow is not None else connection_name
 
                 if (
                     connection_lookup_key(connection_name) != connection_lookup_key(self.default_connection_name)
@@ -194,7 +201,7 @@ class OauthHandlers:
                         error_type = APP_OAUTH_ERROR_TYPES.http_error
                         span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_error_type, error_type)
                         record_exception(span, e)
-                        record_oauth_error(connection_name, APP_OAUTH_OPERATIONS.token_exchange, error_type)
+                        record_oauth_error(event_connection_name, APP_OAUTH_OPERATIONS.token_exchange, error_type)
                         status = status or 500
                         result = APP_OAUTH_RESULTS.failure
                         span.set_attribute(APP_ATTRIBUTE_NAMES.invoke_response_status, status)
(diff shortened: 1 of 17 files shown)


packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -46,13 +69,66 @@ def __init__(
...
+        # which is what stops two concurrent duplicates from both starting an exchange.
+        #
+        # In-flight is checked first so a running exchange always wins over its own
+        # completed marker. The marker is stamped the moment the token is redeemed,
+        # while the owner still has sign-in callbacks to run, and a duplicate that
+        # landed in that window must mirror the owner rather than answer ahead of it.
+        in_flight = self._token_exchange_in_flight.get(exchange_id)
+        if in_flight is not None:
+            return await self._await_token_exchange(ctx, exchange_id, in_flight)
+        if self._is_completed_token_exchange(ctx, exchange_id):
```
corinagum, line 109: Suggest moving this check inside `_run_token_exchange`'s `try`, or hoisting the guard so a state read can't escape it. If reading `ctx.state` throws, the `next_handler()` in `finally` will fail.

Same as the comment I left on #573 at `app_oauth.py:280` and `:200`

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -167,6 +248,136 @@ async def sign_in_token_exchange(
...
+            return
+
+        self._prune_completed_token_exchanges()
+        self._token_exchange_completed[exchange_id] = time()
+        while len(self._token_exchange_completed) > _TOKEN_EXCHANGE_DEDUP_MAX_ENTRIES:
+            del self._token_exchange_completed[next(iter(self._token_exchange_completed))]
+        # Persisted too, so a duplicate handled by another process instance still sees
+        # it. Best-effort only: state has no compare-and-set, so the in-memory layer
+        # above remains the authoritative same-instance guard.
+        record_completed_token_exchange(ctx.state, exchange_id)
```
corinagum: Maybe flush the marker mid-turn the way `ctx.sign_in()` flushes its pending hint. A duplicate on another instance while the owner is still running loads a snapshot without the marker and re-runs.
#573 added a mid-turn `ctx.state._save()` to `sign_in()` for this exact race, reasoning that a token-exchange callback can arrive before the turn that sent the card finished.

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -167,6 +248,136 @@ async def sign_in_token_exchange(
...
+            ) as span:
+                span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_connection, connection_name)
+                span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_operation, APP_OAUTH_OPERATIONS.token_exchange)
+                # Shielded: cancelling this waiter must not cancel the future the
+                # owning request still has to resolve.
+                outcome = await asyncio.shield(in_flight)
+                if outcome.error is not None:
+                    result = APP_OAUTH_RESULTS.failure
+                    span.set_attribute(APP_ATTRIBUTE_NAMES.oauth_result, result)
+                    raise outcome.error
```
corinagum: What about only have an exception here and then a cancelled owner becomes a `412`? Apparently in asyncio `CancelledError` is how to signal shutdown, so re-raising one the task never received makes the task misreport itself and can trip `except asyncio.CancelledError` cleanup. A second, smaller thing: replaying one exception instance to several waiters means they all append frames to the same `__traceback__`.

packages/apps/src/microsoft_teams/apps/app_oauth.py
```diff
@@ -167,6 +248,136 @@ async def sign_in_token_exchange(
...
+        if not has_completed_token_exchange(ctx.state, exchange_id):
+            record_completed_token_exchange(ctx.state, exchange_id)
+
+    def _replay_completed_token_exchange(
+        self, ctx: ActivityContext[SignInTokenExchangeInvokeActivity], exchange_id: str
+    ) -> TokenExchangeResult:
+        """Answer a duplicate that arrived after its exchange already completed."""
+        logger.debug("Duplicate signin/tokenExchange with id '%s' - returning 200 no-op.", exchange_id)
+        self._stamp_completed_token_exchange(ctx, exchange_id)
+        connection_name = ctx.activity.value.connection_name
```
corinagum: Suggest resolving through the registry here and at line 345, the way `_run_token_exchange` does with `event_connection_name`.

MehakBindra (approved):

PR #578, opened 2026-08-27
fix: include typing markers in package wheels

### LINKED ISSUE #560: [Bug]: packages are missing `py.typed` files
### Bug Description

a `py.typed` file is required otherwise type checkers will treat the package as untyped. see https://typing.python.org/en/latest/spec/distributing.html#packaging-type-information

pyright for example complains on every import:

### Steps to Reproduce

```py
from microsoft_teams.cards import AdaptiveCard
```

### Expected Behavior

no type error

### Actual Behavior

pyright error:
```
Stub file not found for "microsoft_teams.cards" (reportMissingTypeStubs)
```

### SDK Version

2.0.15

### Python Version

3.14.6

### Additional Context

_No response_

russlan23:
Fixes #560

## What changed

Adds `py.typed` to each of the seven published `microsoft_teams` distributions and removes the obsolete ignore rule that prevented these markers from being tracked.

The regression test builds every workspace wheel and verifies that each artifact contains its package marker. Runtime behavior is unchanged.

## Validation

- `pytest packages` — 1101 passed
- `ruff format --check` — clean
- `ruff check` — clean
- `pyright` — 0 errors, 0 warnings
- Echo example startup/shutdown smoke test — passed


russlan23:
@microsoft-github-policy-service agree

lilyydu (approved):
LGTM. Thanks for making the change! 

The only wrinkle: #560 asks for "no type error" on `from microsoft_teams.cards import AdaptiveCard`, and this will change it from`reportMissingTypeStubs` to `reportPrivateImportUsage` rather than clearing it. This is a separate issue though, moreso b/c of the generated file - will look into this internally.

Issue #579, opened 2026-08-27
[Bug]: ValidationError on installationUpdate upgrade


jerbob92:
### Bug Description

```
Input tag 'upgrade' found using 'action' does not match any of the expected tags: 'add', 'remove'
```

### Steps to Reproduce

No idea how to get this event.

### Expected Behavior

No validation error

### Actual Behavior

Validation error

### SDK Version

v2.0.16

### Python Version

3.14.7

### Additional Context

Payload data:

```
{
  "action": "upgrade",
  "channelId": "msteams",
  "conversation": {
    "conversationType": "personal",
    "id": "xxx",
    "tenantId": "xxx"
  },
  "entities": [
    {
      "locale": "en-US",
      "type": "clientInfo"
    }
  ],
  "from": {
    "aadObjectId": "xxx",
    "id": "xxx"
  },
  "id": "xxx",
  "recipient": {
    "id": "xxx",
    "name": "xxx"
  },
  "serviceUrl": "https://smba.trafficmanager.net/emea/xxx/",
  "timestamp": "2026-08-26T13:38:36.356Z",
  "type": "installationUpdate"
}
```

PR #583, opened 2026-08-29
fix: accept installationUpdate upgrade action

jibin7jose:
## Summary
Fixes issue #579 

where `installationUpdate` payloads with `action = "upgrade"` raised a validation error because only `add` and `remove` were accepted.

## Changes
- Add support for `installationUpdate.action = "upgrade"`
- Extend the install update activity union with `UpgradedActivity`
- Add a regression test for the upgrade payload from issue #579

## Validation
- `packages/api/tests/unit/test_install_update.py`

### DIFF
--- a/packages/api/src/microsoft_teams/api/activities/install_update/upgrade.py
+++ b/packages/api/src/microsoft_teams/api/activities/install_update/upgrade.py
@@ -1,0 +1,15 @@
+"""
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from typing import Literal
+
+from ...models import ActivityBase, CustomBaseModel
+
+
+class UpgradedActivity(ActivityBase, CustomBaseModel):
+    type: Literal["installationUpdate"] = "installationUpdate"  #
+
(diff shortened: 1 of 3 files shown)


packages/api/src/microsoft_teams/api/activities/install_update/upgrade.py
```diff
@@ -0,0 +1,15 @@
...
+Copyright (c) Microsoft Corporation. All rights reserved.
+Licensed under the MIT License.
+"""
+
+from typing import Literal
+
+from ...models import ActivityBase, CustomBaseModel
+
+
+class UpgradedActivity(ActivityBase, CustomBaseModel):
```
lilyydu: nit: let's name to `InstalledUpgradeActivity`

lilyydu (commented):
Thanks for tackling this @jibin7jose! Just a small nit

packages/api/src/microsoft_teams/api/activities/install_update/upgrade.py
jibin7jose: Updated the class name and fixed the imports/tests.

lilyydu (approved):

PR #585, opened 2026-09-01
fix: use public httpx Request/Response and bound httpx below 1.0

### LINKED ISSUE #568: httpx dependency is unbounded while the SDK imports httpx private modules, so httpx 1.0 will break installs
## Summary

`microsoft-teams-common` imports from httpx's **private** modules while declaring an **unbounded** httpx requirement. httpx 1.0 removes those modules, so once 1.0 is released as stable, a fresh `pip install microsoft-teams-apps` will resolve to httpx 1.0 and fail at import time.

This is not yet affecting users: current httpx stable is 0.28.1, and dev releases aren't selected by default resolvers. Filing it now because httpx 1.0 is actively moving (`1.0.dev4` was published 2026-08-19) and this breaks the moment it ships.

## The two halves of the problem

**1. Private API imports** in `packages/common/src/microsoft_teams/common/http/client.py`:

```python
from httpx._models import Request, Response
from httpx._types import QueryParamTypes, RequestContent, RequestData, RequestFiles
```

**2. No upper bound** on the requirement:

| File | Constraint |
|---|---|
| `packages/common/pyproject.toml` | `httpx>=0.28.1` |
| `packages/apps/pyproject.toml` | `httpx>=0.27.0` |
| `examples/a2a/pyproject.toml` | `httpx>=0.27` |

Either alone would be survivable. Together they mean the next httpx major silently becomes an install-time failure.

## Reproduction

```
$ uv pip install --prerelease=allow "httpx==1.0.dev4" "microsoft-teams-api==2.1.0a2"
$ python -c "import microsoft_teams.api"
...
  File ".../microsoft_teams/common/http/client.py", line 13, in <module>
    from httpx._models import Request, Response
ModuleNotFoundError: No module named 'httpx._models'
```

I hit this accidentally while verifying the 2.1.0a2 release: `--prerelease=allow` applied globally, pulled httpx `1.0.dev4`, and every `import microsoft_teams.*` failed.

## Why 1.0 breaks it

httpx 1.0 is a significant restructure. `_models.py` and `_types.py` no longer exist; the module layout is now `_request.py`, `_response.py`, `_client.py`, `_content.py`, and friends.

```
httpx 0.28.1 : httpx._models OK, httpx._types OK
httpx 1.0.dev4: httpx._models -> ModuleNotFoundError
                httpx._types  -> ModuleNotFoundError
```

## Fixing the imports

The two lines are not equally easy.

`Request` and `Response` are public in both 0.x and 1.0, so that import has a direct public replacement:

```python
from httpx import Request, Response
```

The `_types` aliases (`QueryParamTypes`, `RequestContent`, `RequestData`, `RequestFiles`) are the harder half. They are **not** exported publicly in 0.x or 1.0, so there is no drop-in swap. Options: define our own type aliases in `common/http`, or loosen those annotations. Worth a deliberate decision rather than a mechanical edit, since they appear in public method signatures.

## Suggested actions

1. Switch `Request`/`Response` to the public `from httpx import ...` (safe, works on both).
2. Decide how to replace the `_types` aliases, most likely SDK-owned aliases in `common/http`.
3. Add an upper bound (`httpx>=0.28.1,<1.0`) as a stopgap so a future httpx release can't break installs before 1 and 2 land.

Doing 3 alone is enough to stop the bleeding, but it shouldn't be the permanent answer: the underlying issue is depending on private API.

## Affected

`main`, `release/v2.0` (2.0.16), and `release/v2.1` (2.1.0a2) all carry both private imports and the unbounded constraint.

lukiod:
Fixes #568.

httpx 1.0 removes `httpx._models` and `httpx._types`, and the requirement was unbounded, so once 1.0 ships stable a fresh install resolves to it and every `import microsoft_teams.*` fails at import time.

This does the two safe halves of the issue:

- `Request` and `Response` are public in both 0.x and 1.0, so `packages/common/.../http/client.py` now imports them from `httpx` directly.
- `<1.0` upper bound added in `packages/common`, `packages/apps` and `examples/a2a` (plus the matching `uv.lock` lines).

Left alone deliberately: the `_types` aliases (`QueryParamTypes`, `RequestContent`, `RequestData`, `RequestFiles`). They have no public equivalent in either 0.x or 1.0, and they appear in a lot of public method signatures, so picking a replacement is an API decision rather than a mechanical edit — the upper bound covers them until you decide. Happy to follow up with SDK-owned aliases if you want to go that way.

Verified locally on httpx 0.28.1:

- `from httpx import Request, Response` resolves, and `microsoft_teams.common.http.client` still imports.
- `uv run ruff check .` — All checks passed
- `uv run pytest -q` — 1340 passed

One note: `ruff format --check` reports `packages/api/src/microsoft_teams/api/activities/install_update/__init__.py` would be reformatted. That is pre-existing on `main` — same result with this branch stashed — and untouched here.

### DIFF
--- a/examples/a2a/pyproject.toml
+++ b/examples/a2a/pyproject.toml
@@ -13,7 +13,7 @@ dependencies = [
     "agent-framework-core",
     "agent-framework-openai",
     "uvicorn>=0.30",
-    "httpx>=0.27",
+    "httpx>=0.27,<1.0",
 ]
 
 [tool.uv.sources]
--- a/packages/apps/pyproject.toml
+++ b/packages/apps/pyproject.toml
@@ -32,7 +32,7 @@ graph = [
 test = [
     "pytest>=8.0.0",
     "pytest-asyncio>=0.24.0",
-    "httpx>=0.27.0",
+    "httpx>=0.27.0,<1.0",
 ]
 
 [build-system]
--- a/packages/common/pyproject.toml
+++ b/packages/common/pyproject.toml
@@ -11,7 +11,7 @@ authors = [
 ]
 requires-python = ">=3.11,<4.0"
 dependencies = [
-    "httpx>=0.28.1",
+    "httpx>=0.28.1,<1.0",
 ]
 
 [dependency-groups]
--- a/packages/common/src/microsoft_teams/common/http/client.py
+++ b/packages/common/src/microsoft_teams/common/http/client.py
@@ -10,7 +10,7 @@ from dataclasses import dataclass, field, replace
 from typing import Any, Awaitable, Callable, Dict, List, Optional, Protocol
 
 import httpx
-from httpx._models import Request, Response
+from httpx import Request, Response
 from httpx._types import QueryParamTypes, RequestContent, RequestData, RequestFiles
 
 from .client_token import Token, resolve_token
--- a/uv.lock
+++ b/uv.lock
@@ -84,7 +84,7 @@ requires-dist = [
     { name = "agent-framework-core" },
     { name = "agent-framework-openai" },
     { name = "dotenv", specifier = ">=0.9.9" },
-    { name = "httpx", specifier = ">=0.27" },
+    { name = "httpx", specifier = ">=0.27,<1.0" },
     { name = "microsoft-teams-apps", editable = "packages/apps" },
     { name = "microsoft-teams-cards", editable = "packages/cards" },
     { name = "microsoft-teams-common", editable = "packages/common" },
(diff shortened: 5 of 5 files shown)


lilyydu:
Hey @lukiod,

Thanks for filing this. We'll ideally want to fix the private _types aliases in the same PR as well but need to discuss how we want to handle this with the team first. It'll take some time before we reach a decision, just letting you know!

lukiod:
> Hey @lukiod,
> 
> Thanks for filing this. We'll ideally want to fix the private _types aliases in the same PR as well but need to discuss how we want to handle this with the team first. It'll take some time before we reach a decision, just letting you know!

Sounds good

lilyydu:
Hey @lukiod,

Going to close this ticket as our team is now re-considering a complete swap over the client library. HTTPX is now a supply chain risk as the maintainer has shut the project down completely without clear explanation/communication. We will be looking at alternative solutions instead.

Thanks for contributing, however!

PR #586, opened 2026-09-01
Route threaded replies through Bot Framework reply endpoint

MehakBindra:
## Summary
- route L2 group-chat and channel placement through Bot Framework reply endpoints while retaining L1 sends
- add explicit regular and targeted conversation reply APIs; targeted L2 replies use `POST /v3/conversations/{conversationId}/activities/{rootMessageId}?isTargetedActivity=true`
- infer targeted delivery from outbound recipient metadata in proactive `App.reply()` and reactive `ActivityContext.send()`, including targeted activities placed in existing threads
- preserve proactive callers by translating valid legacy `;messageid=` conversation IDs internally
- expose typed, read-only inbound `channelData.thread.id` and apply the reactive personal/group-chat/channel scope matrix
- keep placement independent from explicit `MessageActivityInput.add_quote()` metadata and preserve quote payloads on targeted replies
- omit transport-only `from` and `conversation` fields from outbound payloads, matching .NET and preventing targeted quoted channel replies from being placed at L1
- accept successful reply responses that omit an activity ID, as observed from APX for regular L2 replies
- update the interacting-with-messages sample with default send, proactive thread, quote, targeted, and targeted+quote cases

## Compatibility
Existing regular reply APIs and public send/reply signatures remain supported. Legacy threaded conversation IDs are translated before reaching APX, and the legacy constructor plus quote/reply conveniences are deprecated without changing runtime behavior. Personal-chat targeted restrictions remain unchanged. Transport routing remains in the endpoint/conversation reference rather than duplicated in serialized activity payloads.

## Validation
- 1,190 API/apps tests passed (wheel build test exercised separately by the successful package build)
- focused targeted/reply/outbound payload suites passed
- `uv run poe check`
- `uv run pyright`
- `uv build --all-packages`
- echo sample startup smoke test
- live channel validation: proactive targeted + quote remains in the selected thread

### DIFF
--- a/examples/interacting-with-messages/src/threading_handlers.py
+++ b/examples/interacting-with-messages/src/threading_handlers.py
@@ -3,7 +3,7 @@ Copyright (c) Microsoft Corporation. All rights reserved.
 Licensed under the MIT License.
 """
 
-from microsoft_teams.api import MessageActivity
+from microsoft_teams.api import MessageActivity, MessageActivityInput
 from microsoft_teams.apps import ActivityContext, App
 
 
@@ -11,7 +11,7 @@ async def handle_thread_reply(ctx: ActivityContext[MessageActivity], text: str) 
     """Send a reactive threaded reply when the command matches."""
     if text != "thread reply":
         return False
-    await ctx.reply("This is a threaded reply to your message.")
+    await ctx.send(MessageActivityInput().add_quote(ctx.activity.id, "This is a threaded reply to your message."))
     return True
 
 
--- a/examples/interacting-with-messages/src/threading_handlers.py
+++ b/examples/interacting-with-messages/src/threading_handlers.py
@@ -5,6 +5,7 @@ Licensed under the MIT License.
 
 from microsoft_teams.api import MessageActivity, MessageActivityInput
 from microsoft_teams.apps import ActivityContext, App
+from microsoft_teams.apps.utils.thread import parse_threaded_conversation_id
 
 
 async def handle_thread_reply(ctx: ActivityContext[MessageActivity], text: str) -> bool:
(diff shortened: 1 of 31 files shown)


packages/apps/src/microsoft_teams/apps/app.py
```diff
@@ -425,8 +425,11 @@ async def reply(  # type: ignore[reportInconsistentOverload]
         if activity is not None:
             if not isinstance(message_id, str):
                 raise TypeError("message_id must be a string when activity is provided")
-            return await self.send(
-                to_threaded_conversation_id(conversation_id, message_id),
+            if not message_id or not message_id.isdigit() or message_id == "0":
+                raise ValueError(f'Invalid message_id "{message_id}": must be a non-zero numeric value')
```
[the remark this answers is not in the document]
MehakBindra: This validation intentionally preserves the behavior previously provided by `to_threaded_conversation_id()`, now that `App.reply()` no longer routes through that deprecated helper. It also keeps legacy suffixed IDs passed to `App.send()` compatible with the same validation. Supporting arbitrary non-empty thread root IDs would be a separate cross-SDK behavior change, since the equivalent high-level APIs in .NET and TypeScript retain numeric validation.

examples/interacting-with-messages/src/threading_handlers.py
```diff
@@ -33,16 +33,15 @@ async def handle_proactive_thread(app: App, ctx: ActivityContext[MessageActivity
...
     if text != "thread manual":
         return False
     conversation_id, thread_root_id = _thread_reference(ctx)
-    thread_id = to_threaded_conversation_id(conversation_id, thread_root_id)
-    await app.send(thread_id, "This was sent using to_threaded_conversation_id() + app.send() for manual control.")
+    await app.reply(conversation_id, thread_root_id, "This was sent using app.reply() with an explicit thread root.")
     return True
 
 
 def _thread_reference(ctx: ActivityContext[MessageActivity]) -> tuple[str, str]:
```
lilyydu: should we consider offering this as apart of the SDK? vs having them write a helper fn

packages/apps/src/microsoft_teams/apps/utils/thread.py
```diff
@@ -25,3 +36,38 @@ def to_threaded_conversation_id(conversation_id: str, message_id: str) -> str:
...
+    root message, the inbound activity ID is the thread root.
+    """
+    conversation_id, legacy_thread_root_id = parse_threaded_conversation_id(activity.conversation.id)
+    thread = activity.channel_data.thread if activity.channel_data else None
+    thread_root_id = thread.id if thread and thread.id else legacy_thread_root_id or activity.id
+    return conversation_id, thread_root_id
+
+
+def get_default_thread_id(activity: MessageActivity) -> str | None:
+    """Resolve the thread root for a reactive send in the inbound scope."""
```
lilyydu, line 62: should we ever fall back to `activity.reply_to_id` in this method?

PR #590, opened 2026-09-08
refactor(apps): add transport seam at HttpServerAdapter layer

lilyydu:
Port of microsoft/teams.ts#720. 

Adds a transport-agnostic inbound seam so non-HTTP transports (e.g. a future Socket Mode adapter) can plug into HttpServer.

- HttpRequest.token: transports that authenticate at the connection level (e.g. Socket Mode) can deliver a pre-resolved token; HttpServer trusts it and skips per-request JWT validation. HTTP adapters never set it, so HTTP behavior is unchanged.
- `HttpServerInitializeDeps` + optional `adapter.initialize(deps)` hook: `HttpServer.initialize()` forwards credentials/cloud to the adapter via getattr(), so it isn't a required Protocol member (avoids breaking existing/external adapters that don't implement it.
- Kept `HttpServer.initialize()` sync (TS's is async) since the actual Socket Mode `adapter.initialize()` does no awaiting.

Pure refactor: no behavior change for existing HTTP adapters.


packages/apps/src/microsoft_teams/apps/http/http_server.py
```diff
@@ -115,6 +115,14 @@ def initialize(
             )
 
         self._adapter.register_route("POST", self._messaging_endpoint, self.handle_request)
+
+        # Forward app-level deps to the adapter so connection-authenticated transports
+        # (e.g. Socket Mode) can initialize through the seam. Optional: most HTTP
+        # adapters (e.g. FastAPIAdapter) don't implement this hook.
+        adapter_initialize = getattr(self._adapter, "initialize", None)
```
heyitsaamir: avoid getattrs. use typed values instead. (i thought i had this in agents.md :()

lilyydu:
closing due to re-design! these changes are no longer needed.

PR #593, opened 2026-09-10
fix: return stream ID when closing streamed activity

### LINKED ISSUE #592: [Bug]: ctx.stream.close() returns placeholder activity ID
### Bug Description

`await ctx.stream.close()` returns a `SentActivity` whose `id` is
`DO_NOT_USE_PLACEHOLDER_ID` instead of the ID of the streamed activity.

Teams returns the stream/activity ID from the initial streaming request. Later streaming requests, including the final request, return an empty successful response. The SDK currently maps the empty response to its placeholder ID and returns that value from `HttpStream.close()`, even though the stream has stored the actual ID from the initial request.

### Steps to Reproduce

1. Emit content through `ctx.stream.emit()`.
2. Await the activity with `ctx.stream.close()`.
3. Observe that `result.id` is `DO_NOT_USE_PLACEHOLDER_ID`.

### Expected Behavior

`ctx.stream.close()` should return a `SentActivity` with the original stream ID, or at least not the `DO_NOT_USE_PLACEHOLDER_ID` value.

### Actual Behavior

`ctx.stream.close()` returns a `SentActivity` set to `DO_NOT_USE_PLACEHOLDER_ID`.

### SDK Version

v2.0.0a31

### Python Version

3.12.10

### Additional Context

_No response_

AlxBit:
Fixes #592

## Summary

`ctx.stream.close()` now returns the original stream activity ID instead of `DO_NOT_USE_PLACEHOLDER_ID` when Teams returns an empty response for the final streaming request.

The stream already retains the ID returned by the initial streaming request. This change uses that retained ID for the final `SentActivity`, including the value emitted to `on_close` handlers and cached for repeated `close()` calls.

## Tests

- Added regression coverage for a final streaming response with no ID.
- Verified that `close()` and the close event handler both receive the original
  stream ID.
- Run:

  ```bash
  uv run --python python3 pytest -q packages/apps/tests/test_http_stream.py
  ```


AlxBit:
@microsoft-github-policy-service agree

corinagum (changes requested):
@AlxBit Thanks for filing a PR, and apologies for the delay in getting back to you! I think in the end this looks correct, but the fix is a layer too late and not fixing `on_chunk`. 

The root cause is `SentActivity.merge` in `sent_activity.py:21`. `{**activity_params.model_dump(), **curr_activity.model_dump()}` runs, and because `SentActivity.id` is required and the api client fills it with `DO_NOT_USE_PLACEHOLDER_ID` on an empty body, the response id always clobbers the request id. Worth noting TS never hits this: `send()` there returns `{ ...payload, ...res }`, so an empty `res` leaves `payload.id` intact. 

Because the fix is in `close()`, the chunk path is unfixed.  Chunks 2 - N also route through `create_activity` (the `streaminfo` entity forces the create branch), also get an empty body, and still emit the placeholder at `http_stream.py:382`. That isn't internal: `app_process.py:226` forwards `on_chunk` into `event_manager.on_activity_sent`, so every plugin sees `DO_NOT_USE_PLACEHOLDER_ID` as the activity id for most of a stream. 

Could you make the correction in `_send` instead?
```
if to_send.id:
    res = res.model_copy(update={"id": to_send.id})
return SentActivity.merge(to_send, res)
```

That covers the final activity, the chunk events, and any future caller in one place, and it reproduces the TS semantics rather than special-casing one call site. I tried it locally with the `close()` hunk reverted and the full apps suite passes.

Btw, the sentinel is hardcoded rather than imported, and the real one (`_PLACEHOLDER_ACTIVITY_ID`, `activity.py:24`) is private and unexported. @LilyDu what are your thoughts? the value is ``"DO_NOT_USE_PLACEHOLDER_ID"``, I feel it's ok not to include a `and res.id == PLACEHOLDER_ACTIVITY_ID:` check.

Lastly, I think the code at `test_http_stream.py:48` could use updating. it returns a fresh real id on every call, which which doesn't match the api client, so no unit tests would catch it. Could we make it return an id on the first call and the placeholder afterwards to reproduce the client's behavior? It makes this whole class of bug fail by default instead of needing a mock each time, and it makes the override in `test_close_waits_for_flush_to_complete` unneeded.

AlxBit:
Thanks for the review @corinagum ! I've moved the fix to `_send()` instead and omitted the placeholder check. I agree that it's okay to not include it. I originally added it after Copilot suggested it.

> Lastly, I think the code at `test_http_stream.py:48` could use updating. it returns a fresh real id on every call, which which doesn't match the api client, so no unit tests would catch it. Could we make it return an id on the first call and the placeholder afterwards to reproduce the client's behavior? It makes this whole class of bug fail by default instead of needing a mock each time, and it makes the override in `test_close_waits_for_flush_to_complete` unneeded.

I've also updated the `mock_send()` in the fixture to mimic the client's behavior.

lilyydu (approved):
LGTM!!

corinagum (approved):

PR #594, opened 2026-09-10
feat(apps): resolve Agentic User file bytes through Graph /shares

corinagum:
### What this does

Agentic User file attachments arrive with a `contentUrl` and **no** `downloadUrl`, so the pre-authorized ("tempauth") URL traditional bots rely on is absent. This resolves those bytes through Microsoft Graph, `GET /shares/{u!encoded}/driveItem/content`, using the Agentic User's own credential.

The route is chosen from the **shape of the attachment**, not from the identity: a `contentUrl` with no `downloadUrl`, at personal scope, resolves via `graphShare`. Everything else keeps the existing `preauthUrl` path unchanged.

### What it deliberately does not do

**An expired pre-authorized URL is terminal.** It raises `FileUrlExpiredError` and makes no Graph call.

A developer can still call Graph with their own client, and keeps the locator because `contentUrl` is public API. The constraint is on what the SDK does automatically.

Auditing trap: the agentic token is itself delegated-shaped, carrying `scp` rather than `roles`. Follow the identity, not the word "delegated".

### Why `.default` and not a named scope

A by-name request hard-fails on exact-string consent mismatch: a blueprint consented to `Files.ReadWrite.All` returns `AADSTS65001` and **no token at all** for a `Files.Read.All` request. Graph's own `/shares` docs list `Files.ReadWrite.All` as least-privileged and list no read-only permission, so admins following Microsoft's guidance land exactly where a by-name request breaks.

The cost is a late, ambiguous 403 when a file permission is missing. We pay that in the SDK instead: the pre-flight rejects a token carrying permissions but **none file-capable**, which restores an early, legible failure. A `Files.Read.All`-then-`.default` ladder was rejected, because `AADSTS65001` is also what "no file permission consented at all" produces, so the ladder makes the most common deployment error harder to diagnose, not easier.

**For security review:** under `.default` the token carries the blueprint's full consented set, including unrelated scopes. The mitigation is the grant, not the request string: do not consent unrelated scopes on a file-handling blueprint.

### Testing

Unit and integration coverage is in the diff. All three scenarios were also run **live against real Teams**, tenant asserted twice per run (browser session state, and the SharePoint host in the bot's own log).

| Scenario | Result |
| --- | --- |
| Traditional bot, fresh file | `preauthUrl`, byte-exact |
| Agentic User via Graph | `/shares/{u!}/driveItem/content` 200, byte-exact, driving the shipped helpers |
| Expired URL | `FileUrlExpiredError`, **zero requests to `graph.microsoft.com` in the entire process log** |

The third row was checked by grepping the whole bot process log for an *absence*, not by reading the rendered result, and the expiry is real rather than simulated: the same `UniqueId` returns 200 early and a hard 401 later. And the pre-authorized URL lifetime is **not a constant**: three runs bounded the flip at (62.7, 72.4], (62.5, 68.4] and (63.6, 66.4] minutes. Poll rather than picking a wait time.

---

## This language

20 commits, 25 files, +2022 / -67.

**`Follow storage redirects on the pre-authorized download path`** closes a real divergence rather than introducing one. `httpx` does not follow redirects by default, unlike the TS and .NET transports, so a storage 302 surfaced as a bare `RuntimeError` instead of bytes. Pinned by unit tests and transport parity, and confirmed live on 2026-09-11: Graph answered `302` to `tsdkjuly2026-my.sharepoint.com` and the walk followed it to a `200`.


packages/apps/src/microsoft_teams/apps/files/download.py
```diff
@@ -47,7 +86,8 @@ async def open_file_stream(
     *,
     prior_fetch_succeeded: bool = False,
     client: Optional[httpx.AsyncClient] = None,
-) -> AsyncIterator[OpenedFileStream]:
+    credential: Optional[GraphCredential] = None,
+) -> AsyncGenerator[OpenedFileStream]:
```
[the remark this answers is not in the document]
corinagum: This is incorrect: `collections.abc.AsyncGenerator[X]` builds `GenericAlias` and there are no `TypeError`s.

packages/apps/src/microsoft_teams/apps/files/incoming_file.py
```diff
@@ -138,11 +143,12 @@ async def save_as(self, path: str) -> None:
                 await asyncio.to_thread(file.close)
 
     @asynccontextmanager
-    async def _open(self) -> AsyncIterator[OpenedFileStream]:
+    async def _open(self) -> AsyncGenerator[OpenedFileStream]:
```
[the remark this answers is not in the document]
corinagum: This is incorrect: `collections.abc.AsyncGenerator[X]` builds `GenericAlias` and there are no `TypeError`s.

examples/ai-file-analysis/README.md
```diff
@@ -74,6 +74,27 @@ file round-trip is demonstrable without a model subscription. Only the analysis
...
+- **A bot reads through the pre-authorized download URL.** Teams provides a URL on the attachment and the SDK fetches it directly, with no tokens attached, shown in this example.
+- **An agentic user reads through Microsoft Graph.** Agentic Users never receive a pre-authorized URL, so the SDK resolves the file through Graph's `/shares` endpoint using the agent's own credential.
+
+The route is determined by whether or not the pre-authorized URL exists. If it does, it is used. Otherwise, the Graph path is followed.
+
+### The download URL expires, and that is final
+
+A pre-authorized URL is short-lived, on the order of an hour, and the exact lifetime is set per URL rather than fixed. Do not depend on a particular number. Once it lapses `download()` raises `FileUrlExpiredError` and those bytes are unreachable: there is no recovery, and the file has to be sent again.
+
+That shapes how you write a handler rather than being a detail to note. Download when the file arrives, as this sample does, instead of storing an `IncomingFile` to read later. Queued work, retry-with-backoff, and anything that waits on a person are the shapes that run into it.
```
lilyydu: can we rephrase this last part? I'm not sure I fully understand it

"Queued work, retry-with-backoff, and anything that waits on a person are the shapes that run into it."

examples/ai-file-analysis/README.md
```diff
@@ -74,6 +74,27 @@ file round-trip is demonstrable without a model subscription. Only the analysis
...
+
+### Agentic users
+
+The handler code is the same: `list()` then `download()`, with the SDK selecting the agent's own credential rather than the app's. A failed Graph read raises `FileRetrievalError`, which names the identity that was refused and links the permissions documentation for it.
+
+Running it is what differs, and sideloading this sample will not get you there. An agentic user is published through the Teams admin center rather than an app manifest, receives activities on its blueprint's notification URL rather than at your bot endpoint, and needs a Graph file permission consented on that blueprint by an administrator. If you already have one provisioned, point its notification URL at this sample and the code runs unchanged. To provision one, start with the [Microsoft Agent 365 documentation](https://learn.microsoft.com/microsoft-agent-365/), and see [inheritable permissions](https://learn.microsoft.com/entra/agent-id/concept-inheritable-permissions) for how an agent gets the Graph scopes this path needs.
```
lilyydu: should we put this before the first paragraph? And maybe condense it, the first sentence and general tone of the README is a bit overly wordy

examples/ai-file-analysis/README.md
corinagum: Sure, will fix to "Anything that delays the read can outlive the URL: a job queue, a retry with backoff, or a step that waits on a person."

packages/apps/src/microsoft_teams/apps/files/download.py
```diff
@@ -105,6 +175,180 @@ async def _open_personal_file_stream(
...
+        # because that URL carries its own credential, this one requires it.
+        request = http.build_request("GET", url, headers={"Authorization": f"Bearer {token}"})
+        response = await http.send(request, stream=True, follow_redirects=True)
+
+        try:
+            if response.status_code in (401, 403):
+                # An unconsented scope, a file never shared with this identity, and a drive item that does not exist
+                # are all 403, differing only in message text. The SDK cannot branch on that, but the developer can
+                # read it, so it is carried rather than dropped.
+                raise FileRetrievalError("access_denied", actor, await _read_service_error(response))
```
lilyydu: should we treat 401 and 403 differently?

packages/apps/src/microsoft_teams/apps/files/errors.py
```diff
@@ -61,3 +88,88 @@ def __init__(self, scope: ConversationType, message: Optional[str] = None) -> No
...
+        return (
+            "the app has no usable Graph credential for this file. Graph file retrieval is supported for Agentic "
+            "Users, which read as their own identity; an app identity and/or user-delegated permissions may be "
+            "used but are not supported via the SDK at this time"
+        )
+    assert_never(actor)
+
+
+def _default_retrieval_message(reason: FileRetrievalFailureReason, actor: Optional[FileActor]) -> str:
+    as_who = _describe_actor(actor or "app")
```
lilyydu: this seems to default to "app" even if actor was None

packages/apps/src/microsoft_teams/apps/files/errors.py
```diff
@@ -61,3 +88,88 @@ def __init__(self, scope: ConversationType, message: Optional[str] = None) -> No
             message = f"downloading files from '{scope}' conversations is not supported via SDK at this time"
         super().__init__(message)
         self.scope = scope
+
+
+class FileRetrievalError(FileError):
```
lilyydu: this feels kinda heavy for two failure cases. could we simplify it to smth like `FileCredentialError` that keeps the original exception, and `FileAccessError` with the status code, actor, and service details?

packages/apps/src/microsoft_teams/apps/app_process.py
```diff
@@ -155,6 +161,17 @@ async def _build_context(
...
 
+        # Resolved at fetch time rather than eagerly, so a turn that never touches files pays nothing for it.
+        #
+        # An Agentic User reads as itself. An app-only token sees what the app may read tenant-wide, a different set
+        # from what was shared with the agent, so it would 403 on exactly the files the agent was given.
+        files_credential = select_files_credential(
+            agentic_identity=activity.recipient.agentic_identity,
+            graph_base_url_root=self.graph_base_url_root,
+            get_app_graph_token=lambda: self.get_app_graph_token(tenant_id),
+            get_agentic_graph_token=self.get_agentic_graph_token,
```
lilyydu: should we pass the activity tenant here too? `AgenticIdentity.tenant_id` is optional, but `_get_agentic_graph_token` falls back to the configured tenant so downloads could request a token for the wrong tenant or no tenant

packages/apps/src/microsoft_teams/apps/files/download.py
```diff
@@ -87,10 +149,18 @@ async def _open_personal_file_stream(
         # would be sent to a third-party storage host.
         request = http.build_request("GET", url)
         request.headers.pop("Authorization", None)
-        response = await http.send(request, stream=True)
+
+        # `follow_redirects` is explicit because httpx defaults it to False. Without it a storage 302 never resolves:
+        # it is not 2xx, so it falls through to the `not response.is_success` arm below and surfaces as
+        # "failed to download file: 302 Found". Nothing leaks across the hop, because the request deliberately
+        # carries no Authorization header, so there is no bearer to withhold from a third-party host.
+        response = await http.send(request, stream=True, follow_redirects=True)
```
lilyydu: could we reject redirects to non-HTTPS destinations? `follow_redirects=True` can follow an HTTPS URL to HTTP

packages/apps/src/microsoft_teams/apps/files/errors.py
corinagum: oop, thanks for the catch!

packages/apps/src/microsoft_teams/apps/files/download.py
```diff
@@ -104,6 +104,43 @@ async def open_file_stream(
         yield opened
 
 
+_MAX_REDIRECTS = 10
+"""Matches httpx's own default ceiling, so guarding the scheme does not also change how many hops are allowed."""
```
lilyydu: I think HTTPX's might be 20, if they have a config value maybe we should ref it in case it changes later on

corinagum: Thanks for the catch; updated and added new tests

lilyydu (approved):

lilyydu (approved):

